PR #1, opened 2024-02-12
Created a tool to split .s files automatically

Yanis42:
I haven't tried it extensively but it should work for most files, this does 3 things:
1. find functions and split them
2. writes the C file and the different .s files inside ``asm/non_matchings/`` for every parsed function
3. update ``obj_files.mk`` to replace ``asm`` by ``src``

---
It works as a command line tool, example: ``./tools/func_splitter.py asm/system.s``:
<details>

</details>

### DIFF
--- a/tools/func_splitter.py
+++ b/tools/func_splitter.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+import argparse
+
+
+def splitAsm(inputFile: str):
+    # get the original asm file's data
+    with open(f"asm/{inputFile}.s", "r") as asmFile:
+        lines = asmFile.readlines()
+        asmFile.seek(0)
+        rawData = asmFile.read()
+
+    # parse functions names and their data
+    foundFunction = False
+    functionName = ""
+    functionData: list[str] = []
+    parsedFunctions: list[tuple[str, list[str]]] = []
+
+    # if the file has functions
+    if ".section .text" in rawData or ".section .init" in rawData:
+        for i, line in enumerate(lines):
+            previousIsNewLine = lines[i - 1] == "\n"
+
+            # if end of a section containing functions
+            if (
+                line.startswith(".section .") 
+                and not line.startswith(".section .text") 
+                and not line.startswith(".section .init")
+                and previousIsNewLine
+            ):
+                break
+
+            # if we found a new function
+            if (
+                not line.startswith("lbl_") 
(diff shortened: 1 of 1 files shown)


tools/func_splitter.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python3
```
cadmic, line 1: Could you make this file executable? `tools/func_splitter.py` is slightly easier to type than `python3 tools/func_splitter.py`, and otherwise this `#!` line has no purpose

tools/func_splitter.py
```diff
@@ -0,0 +1,109 @@
...
+
+import argparse
+import os
+
+
+def splitAsm(inputFile: str):
+    # get the original asm file's data
+    with open(f"asm/{inputFile}.s", "r") as asmFile:
```
cadmic, line 9: my first instinct was to type `python3 tools/func_splitter.py --file asm/system.s` and was suprised that it wanted `system.s` instead. Typing the full path has the small advantage that my shell will tab-complete for me, but up to you

tools/func_splitter.py
```diff
@@ -0,0 +1,109 @@
...
+                cFile.write(cData + f'#pragma GLOBAL_ASM("asm/non_matchings/{inputFile}/{functionName}.s")' + newline)
+
+            with open("obj_files.mk", "w") as linkerFile:
+                linkerFile.write(data.replace(f"/asm/{inputFile}.o", f"/src/{inputFile}.o"))
+
+
+def main():
+    parser = argparse.ArgumentParser(description="asm function splitter (creates C files and updates obj_files.mk)")
+    parser.add_argument(
+        "--file",
```
cadmic, line 93: Should this be a position argument instead of `--file`? It's slightly less typing and I don't think it would be confusing. Again, up to you

cadmic (approved):
looks very useful, I only have nitpicks

tools/func_splitter.py
Yanis42: Oh yeah, I should do that

tools/func_splitter.py
Yanis42: I thought the python3 isn't mandatory when you execute a script with this line, I do it out of habit but I'll have to try

Yanis42: I tried and I had a permission error but I think it's good now? let me know if it doesn't work

PR #3, opened 2024-02-12
Start system.c

Yanis42:
This just split system.s, some get/set funcs should be easy to match


include/system.h
```diff
@@ -0,0 +1,218 @@
+#ifndef _SYSTEM_H
+#define _SYSTEM_H
+
+#include "types.h"
+
+
+// TODO: should be __anon_0x37240
+// probably wrong but size matches (0x88)
+typedef struct System {
```
cadmic, line 9: I'm pretty sure this is the wrong struct, based on the signature of systemEvent in the debug info it should be __anon_0x37240

include/system.h
```diff
@@ -0,0 +1,218 @@
...
+    /* 0x0164 */ int rumbleConfiguration;
+    /* 0x0168 */ int storageDevice;
+    /* 0x016C */ int normalControllerConfig;
+    /* 0x0170 */ int currentControllerConfig;
+} SystemRomConfig; // size = 0x174
+
+
+int systemEvent(System* pSystem, int nEvent, void *pArgument);
+
+char D_800EB300[] = "SYSTEM (N64)";
```
cadmic, line 39: These data definitions should all go in the .c so it's not redefined by any other .c file that includes this header

Yanis42: yeah I just put them there because it takes lots of lines

cadmic: Maybe we can hide them at the bottom of the C file?

Yanis42: I'm not sure, I feel like some functions may need them
i think it's fine for now, we can always improve that later once the file is matched or something

PR #6, opened 2024-02-15
Match some rom.c functions

Yanis42:
Match:
- ``romEvent``
- ``romGetImage``
- ``romSetImage``
- ``romSetCacheSize``
- ``romUpdate``
- ``romCopy``
- ``romCopyImmediate`` (@cadmic)
- ``__romCopyUpdate_Complete``
- ``__romLoadUpdate_Complete``
- ``romLoadRange`` (this one has weird float issues that requires ``romLoadFullOrPart`` to be matched)
- ``__romLoadBlock_CompleteGCN``
- ``__romLoadBlock_Complete``

Not matching but close:
- [``romLoadFullOrPart``](https://decomp.me/scratch/6c62Q)


include/cpu.h
```diff
@@ -3,32 +3,42 @@
 
 #include "types.h"
 
-// size: 0x34
+typedef int (*pfPut8_func)(struct Rom* pROM, u32 nAddress, s8* pData);
```
cadmic, line 6: In this case, `pfPut8_func` is not a variable that's pointer to a function, it's a type name, so I don't think it should have a `pf` prefix. Probably it should be named CamelCase like our other types like Put8Func.

(Thank you for naming these though!)

include/rom.h
```diff
@@ -3,84 +3,109 @@
 
 #include "xlFileGCN.h"
 
+typedef s32 pCallback_func(void);
```
cadmic, line 6: maybe CompletionCallback for symmetry with ProgressCallback? Is that even what this is?

include/rom.h
```diff
@@ -3,84 +3,109 @@
...
-int romEvent(Rom* pROM, int nEvent, void* pArgument);
+int romEvent(Rom* pROM, s32 nEvent, unknownDeviceStruct* pArgument);
 int romGetImage(Rom* pROM, char* acNameFile);
 int romSetImage(Rom* pROM, char* szNameFile);
 int romSetCacheSize(Rom* pROM, int nSize);
 int romUpdate(Rom* pROM);
-int romCopyImmediate(Rom* pROM, void* pTarget, int nOffsetROM, u32 nSize);
-int romCopy(Rom* pROM, void* pTarget, int nOffset, u32 nSize, int (*pCallback)());
+int romCopyImmediate(Rom* pROM, void* pTarget, int nOffsetROM, int nSize);
+s32 romCopy(Rom* pROM, void* pTarget, s32 nOffset, s32 nSize, pCallback_func* pCallback);
```
cadmic, line 86: nit: I know we haven't settled on s32 vs int but it would be nice if `romCopyImmediate` and `romCopy` were consistent here at least

include/rom.h
```diff
@@ -3,84 +3,109 @@
...
+
+static int romPutDebug8(Rom* pROM, u32 nAddress, s8* pData);
+static int romPutDebug16(Rom* pROM, u32 nAddress, s16* pData);
+static int romPutDebug32(Rom* pROM, u32 nAddress, s32* pData);
+static int romPutDebug64(Rom* pROM, u32 nAddress, s64* pData);
+
+static int romGetDebug8(Rom* pROM, u32 nAddress, s8* pData);
+static int romGetDebug16(Rom* pROM, u32 nAddress, s16* pData);
+static int romGetDebug32(Rom* pROM, u32 nAddress, s32* pData);
+static int romGetDebug64(Rom* pROM, u32 nAddress, s64* pData);
```
cadmic, line 109: static functions shouldn't be declared in a header, because they're not exported so other modules can't make use of them. We should probably stick them at the top of rom.c instead with all the other forward declarations

include/cpu.h
```diff
@@ -3,32 +3,42 @@
...
+    /* 0x28 */ pfPut64_func pfPut64;
+    /* 0x2C */ u32 nAddressPhysical0;
+    /* 0x30 */ u32 nAddressPhysical1;
+} __anon_0x3EB4F; // size = 0x34
+
+typedef struct unknownDeviceStruct {
+    /* 0x00 */ int unk;
+    /* 0x00 */ char unk2[0x20];
+    /* 0x24 */ __anon_0x3EB4F* pDevice;
+} unknownDeviceStruct;
```
cadmic, line 36: I kinda think we should confine this to rom.c for now until we're more sure about it (and not use it in rom.h either and leave pHost as a `void*`)

also nit: UnknownDeviceStruct rather than unknownDeviceStruct

include/rom.h
```diff
@@ -3,84 +3,109 @@
...
+    /* 0x10E6C */ int nSizeCacheRAM;
+    /* 0x10E70 */ u8 acHeader[64];
+    /* 0x10EB0 */ u32* anOffsetBlock;
+    /* 0x10EB4 */ int nCountOffsetBlocks;
+    /* 0x10EB8 */ DVDFileInfo fileInfo;
+    /* 0x10EF4 */ int offsetToRom;
+} Rom; // size = 0x10EF8
 
-int romEvent(Rom* pROM, int nEvent, void* pArgument);
+int romEvent(Rom* pROM, s32 nEvent, unknownDeviceStruct* pArgument);
```
cadmic, line 80: leave as `void*` and cast in `romEvent` instead? I'm not sure if `pArgument` is always the same, or it could be different types depending on the event id in `nEvent`

include/rom.h
```diff
@@ -3,84 +3,109 @@
...
 int romGetImage(Rom* pROM, char* acNameFile);
 int romSetImage(Rom* pROM, char* szNameFile);
 int romSetCacheSize(Rom* pROM, int nSize);
 int romUpdate(Rom* pROM);
-int romCopyImmediate(Rom* pROM, void* pTarget, int nOffsetROM, u32 nSize);
-int romCopy(Rom* pROM, void* pTarget, int nOffset, u32 nSize, int (*pCallback)());
+int romCopyImmediate(Rom* pROM, void* pTarget, int nOffsetROM, int nSize);
```
cadmic, line 85: could you comment that `romCopyImmediate` and `romCopy` take an `unsigned int nSize` instead of a signed int in the debug info? I'm not sure what's up with that but I kinda don't want to forget about it

src/rom.c
```diff
@@ -86,19 +88,370 @@ const f32 D_80135FF8 = 639.0f;
...
+    return 1;
+}
+
+s32 romGetImage(Rom* pROM, char* acNameFile) {
+    if (pROM->acNameFile[0] == '\x0') {
+        return 0;
+    }
+
+    if (acNameFile != NULL) {
+        s32 i;
```
cadmic, line 161: I think that we should try to match the local variable names in the debug info whenever possible, so `iName` here? It is awkwardly long though. Ditto on the rest of the functions (e.g. pFile vs file)

src/rom.c
```diff
@@ -86,19 +88,370 @@ const f32 D_80135FF8 = 639.0f;
...
+    }
+
+    pROM->bFlip = bFlip;
+    simulatorDVDOpen(szNameFile, &pROM->fileInfo);
+}
+
+s32 romSetImage(Rom* pROM, char *szNameFile) {
+    tXL_FILE* file;
+    s32 i;
+    int blockCopySize;
```
cadmic, line 192: pFile, iName, nSize

src/rom.c
```diff
@@ -86,19 +88,370 @@ const f32 D_80135FF8 = 639.0f;
...
-#pragma GLOBAL_ASM("asm/non_matchings/rom/romGetImage.s")
+    pROM->nSizeCacheRAM = nSizeCacheRAM;
+    pROM->nCountBlockRAM = nSizeCacheRAM / 0x2000;
 
-#pragma GLOBAL_ASM("asm/non_matchings/rom/romSetImage.s")
+    if (!xlHeapTake(&pROM->pBuffer, nSizeCacheRAM | 0x30000000)) {
+        return 0;
+    }
+
+    pROM->pCacheRAM = (u8* ) pROM->pBuffer;
```
cadmic, line 243: nit: formatting

you can run `clang-format -i include/*.h src/*.c` to format everything

src/rom.c
```diff
@@ -86,19 +88,370 @@ const f32 D_80135FF8 = 639.0f;
...
+        return 0;
+    } else if (!romCacheGame(pROM)) {
+        return 0;
+    }
+
+    pROM->bLoad = 0;
+    return 1;
+}
+
+//! TODO: remove this when the SDK files are present
```
cadmic, line 296: could you move these declarations to the top?

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
     }
 }
 
+#ifndef NON_MATCHING
 #pragma GLOBAL_ASM("asm/non_matchings/rom/romLoadFullOrPart.s")
+#else
+// weird float issue at ``simulatorShowLoad(1, pROM->acNameFile, D_80135FD0);``
+inline int romLoadFullOrPart_Loop(Rom* pROM) {
```
cadmic, line 607: nit: romLoadFullOrPartLoop would better match naming conventions

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
...
+        } else if (!romLoadBlock(pROM, i, iCache, NULL)) {
+            return 0;
+        }
+    }
+
+    return 1;
+}
+
+s32 romLoadFullOrPart(Rom* pROM) {
+    tXL_FILE* file;
```
cadmic, line 630: pFile

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
...
+    pROM->load.bWait = 0;
+    return 1;
+}
 
 #pragma GLOBAL_ASM("asm/non_matchings/rom/romCacheGame.s")
 
 #pragma GLOBAL_ASM("asm/non_matchings/rom/romCacheGame_ZELDA.s")
 
+//! TODO: remove this when ``romLoadFullOrPart`` is matched
+#if 1
```
cadmic, line 732: `#ifndef NON_MATCHING`? or does this not build at all

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
...
+s32 __romLoadBlock_Complete(Rom* pROM) {
+    s32 iBlock;
+
+    if (pROM->bFlip) {
+        u32* panData = (u32*)pROM->load.anData;
+        u32 temp_r0;
+        u32 i;
+
+        for (i = 0; i < ((pROM->load.nSize + 3) >> 2); i++) {
+            // fake?
```
cadmic, line 784: I don't think it is tbh

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
...
+
+    pROM->load.nResult = nResult;
+    pROM->load.bDone = 1;
+}
+
+s32 __romLoadBlock_Complete(Rom* pROM) {
+    s32 iBlock;
+
+    if (pROM->bFlip) {
+        u32* panData = (u32*)pROM->load.anData;
```
cadmic, line 779: nit: I don't think this is a pointer to an array of numbers (that would be `u32**`) so maybe `anData` or `pData`?

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
...
+        u32 temp_r0;
+        u32 i;
+
+        for (i = 0; i < ((pROM->load.nSize + 3) >> 2); i++) {
+            // fake?
+            temp_r0 = *panData;
+            *panData = ((temp_r0 >> 8) & 0xFF00FF) | ((temp_r0 << 8) & 0xFF00FF00);
```
cadmic, line 786:
```suggestion
            *panData = ((temp_r0 >> 8) & 0x00FF00FF) | ((temp_r0 << 8) & 0xFF00FF00);
```

src/rom.c
```diff
@@ -86,19 +88,370 @@ const f32 D_80135FF8 = 639.0f;
...
+    }
+
+    return 1;
+}
+
+inline void romOpen(Rom* pROM, char *szNameFile) {
+    s32 var_r30 = 0;
+    int bFlip;
+
+    if ((pROM->acHeader[0] == '\x37') && (pROM->acHeader[1] == 0x80)) {
```
cadmic, line 175:
```suggestion
    if ((pROM->acHeader[0] == 0x37) && (pROM->acHeader[1] == 0x80)) {
```
for consistency with 0x80?

src/rom.c
```diff
@@ -247,27 +600,207 @@ int romGetPC(Rom* pROM, u64* pnPC) {
...
+        if (pROM->bFlip) {
+            u32* var_r6 = (u32*)pROM->pBuffer;
+            u32 temp_r0;
+            s32 j;
+
+            for (j = 0; j < (((s32)pROM->nSize + 3) >> 2); j++) {
+                // fake?
+                temp_r0 = *var_r6;
+                *var_r6 = ((temp_r0 >> 8) & 0xFF00FF) | ((temp_r0 << 8) & 0xFF00FF00);
+                var_r6 += 1;
```
cadmic, line 700: is the same function inlined into both this as __romLoadBlock_Complete?

include/rom.h
```diff
@@ -3,84 +3,109 @@
...
-    u32 nSize; // 0x28
-    u32 nSizeRead; // 0x2C
-} __anon_0x4D1DA;
+typedef struct RomLoadState {
+    /* 0x00 */ int bWait;
+    /* 0x04 */ int bDone;
+    /* 0x08 */ int nResult;
+    /* 0x0C */ u8* anData;
+    /* 0x10 */ pCallback_func* pCallback;
+    /* 0x14 */ s32 iCache;
```
cadmic, line 46:
```suggestion
    /* 0x14 */ int iCache;
```

I know we haven't settled in s32 vs int but it would be nice to be consistent in this struct at least

src/rom.c
```diff
@@ -86,19 +88,370 @@ const f32 D_80135FF8 = 639.0f;
...
+
+s32 romCopy(Rom* pROM, void* pTarget, s32 nOffset, s32 nSize, pCallback_func* pCallback) {
+    tXL_FILE* file;
+
+    nOffset &= 0x07FFFFFF;
+
+    if (!pROM->nSizeCacheRAM) {
+        if (!xlFileOpen(&file, XLFT_BINARY, pROM->acNameFile)) {
+            return 0;
+        } else if (!xlFileSetPosition(file, nOffset + pROM->offsetToRom)) {
```
cadmic, line 421: not a big deal but I wonder if we should write

```
if (!xlFileOpen(&file, XLFT_BINARY, pROM->acNameFile)) {
    return 0;
}

if (...) { ... }
```
rather than
```
f (!xlFileOpen(&file, XLFT_BINARY, pROM->acNameFile)) {
    return 0;
} else if (...) { ... }
```
to match the style of the other functions

include/rom.h
Yanis42: I'll name it "UnknownCallbackFunc" until we have more information about how it's used

include/rom.h
Yanis42: I did the cast, we can always change that later if it turns out it can be different types

src/rom.c
Yanis42: I think it's weird to some something else than ``i`` for counters but I did the change for consistency

src/rom.c
Yanis42: I made it a make rule to make this easier

src/rom.c
Yanis42: I forgot to change that, I used if 1 because if was shorter when I had to test things

src/rom.c
Yanis42: ``anData`` is more consistent imo

src/rom.c
Yanis42: 😞 I'm sad we can't use the char notation for 0x80, imo it makes more sense to use that as it's supposed to be a string

src/rom.c
Yanis42: I'm gonna try this after the rest of the review, I'm sure it's inlined with the data pointer as a parameter

src/rom.c
Yanis42: yeah I was thinking the same, I don't think we should use else if it's not actually needed, it brings more confusion (and ugly code ngl)

src/rom.c
Yanis42: doesn't seem to match, I didn't investigate that yet

include/rom.h
```diff
@@ -3,82 +3,96 @@
 
 #include "xlFileGCN.h"
 
+typedef s32 UnknownCallbackFunc();
```
cadmic, line 6: As discussed, because C is weird this should probably be

```suggestion
typedef s32 UnknownCallbackFunc(void);
```

same with __romCopyUpdate_Complete() and __romCopyUpdate_Complete()

cadmic (commented):
(not approving bc you said you wanted to include romLoadUpdate and romCopyUpdate too, lmk if you change your mind)

cadmic (approved):

PR #8, opened 2024-02-15
Match some ROM cache functions

cadmic:
I added some crude documentation for some non-obvious struct fields too

### DIFF
--- a/src/rom.c
+++ b/src/rom.c
@@ -112,7 +112,9 @@ static int romGetDebug64(Rom* pROM, u32 nAddress, s64* pData);
 
 s32 __romCopyUpdate_Complete(void);
 s32 __romLoadUpdate_Complete(void);
-void __romLoadBlock_CompleteGCN(long nResult);
+void __romLoadBlock_CompleteGCN(long nResult, DVDFileInfo* fileInfo);
+int romMakeFreeCache(Rom* pROM, int* piCache, RomCacheType eType);
+int romFindOldestBlock(Rom* pROM, int* piBlock, RomCacheType eTypeCache, int whichBlock);
 
 //! TODO: remove this when the SDK files are present
 u32 ARGetDMAStatus(void);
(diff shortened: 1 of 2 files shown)


src/rom.c
```diff
@@ -852,10 +891,177 @@ s32 __romLoadBlock_Complete(Rom* pROM) {
...
+            return 0;
+        }
+
+        nOffsetRAM = iCacheRAM * 0x2000;
+        nOffsetARAM = iCacheARAM * 0x2000;
+        nOffsetARAM += ARGetBaseAddress();
+
+        while (ARGetDMAStatus()) {}
+
+        ARStartDMA(1, (u32)&pROM->pCacheRAM[nOffsetRAM], nOffsetARAM, 0x2000);
```
Yanis42, line 918: I think the 1 in this function should be ``ARAM_DIR_ARAM_TO_MRAM``

src/rom.c
```diff
@@ -852,10 +891,177 @@ s32 __romLoadBlock_Complete(Rom* pROM) {
...
+        nOffsetRAM = iCacheRAM * 0x2000;
+        nOffsetARAM = iCacheARAM * 0x2000;
+        nOffsetARAM += ARGetBaseAddress();
+
+        DCStoreRange(&pROM->pCacheRAM[nOffsetRAM], 0x2000);
+
+        while (ARGetDMAStatus()) {}
 
-#pragma GLOBAL_ASM("asm/non_matchings/rom/romMakeFreeCache.s")
+        ARStartDMA(0, (u32)&pROM->pCacheRAM[nOffsetRAM], nOffsetARAM, 0x2000);
```
Yanis42, line 939: same here except it's ``ARAM_DIR_MRAM_TO_ARAM`` (according to [this](https://github.com/doldecomp/dolsdk2001/blob/e973c42aeb202d843e20006b94f408ce91e367c9/include/dolphin/ar.h#L6) at least)

this one's not defined btw

Yanis42 (commented):
tiny details, otherwise good to go

src/rom.c
Yanis42: oohh 0 is "dma write" and 1 "dma read", should be worth adding a comment about it somewhere

Yanis42: now that I think about it this call was most likely from the macro from ar.h

cadmic: Thanks, I've replaced these with ARStartDMARead/ARStartDMAWrite

Yanis42 (approved):

PR #9, opened 2024-02-16
Remaining rom functions

Yanis42:
doing a draft until I get ``romCopyUpdate`` matched

### DIFF
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -27,6 +27,11 @@ typedef struct __anon_0x3EB4F {
     /* 0x28 */ Put64Func pfPut64;
     /* 0x2C */ u32 nAddressPhysical0;
     /* 0x30 */ u32 nAddressPhysical1;
+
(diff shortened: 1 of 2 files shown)


Yanis42:
I got ``romLoadFullOrPart`` to match so I removed the ifdefs around it (and for ``romLoadRange`` too), though it doesn't build OK (first diff at 0x6B770), the diff script tells the file match though (with ``diff.py -mwo3 romEvent --max-lines 10000``), I suspect the presence of ``D_80135FD0`` to be the issue (as it doesn't seems like a real variable)

I'll let someone else find the cause I have no idea how to debug that

cadmic:
I think it's fine to punt on the data for now, as long as diff.py says the instructions match that's huge progress

Yanis42:
gonna close it so cadmic can include these changes with ``romCacheGame_ZELDA`` which fixes the issues this branch has

PR #10, opened 2024-02-16
Updated the function splitter script

Yanis42:
New options:
- ``--file`` (``-f``) FILE.s or PATH_TO_FILE.s: the current behavior
- ``--group`` (``-g``) GROUP_NAME: splits every file of a specific group
- ``--all`` (``-a``): splits every files from every groups

Also this adds black as the default Python formatter, I think it's better to have nice looking scripts

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -63,9 +63,7 @@ if __name__ == "__main__":
     except ModuleNotFoundError:
         argcomplete = None
 
-    parser = argparse.ArgumentParser(
-        description="Diff MIPS, PPC, AArch64, ARM32, SH2, SH4, or m68k assembly."
-    )
+    parser = argparse.ArgumentParser(description="Diff MIPS, PPC, AArch64, ARM32, SH2, SH4, or m68k assembly.")
 
     start_argument = parser.add_argument(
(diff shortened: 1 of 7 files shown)


cadmic:
Could you make the PR title more descriptive?

To be honest there's so still so much manual work involved in splitting a file for the first time that I'm not sure it's worth it to try to split multiple at once

As for the formatting changes, I think it belongs it it's own PR. FWIW, I don't like re-formatting either diff.py or asm_processor.py since they're derived from other repos and it will make it harder to diff the changes. Meanwhile diff.py and objdump_wrapper.py are both formatted with black in the default style, and I thought the whole philosophy of black is that it should not be configured ("any color you like, as long as it's black")

Yanis42:
I did that based on the discussion I wanted to have yesterday about how we should organise the files, this update makes it so it's creating the .c/.s files in the correct directory

no rush here of course, this can wait (or maybe not even necessary idk, but just in case we need that it's here and ready to use)

Yanis42:
gonna close this, I don't think it's needed anymore (now that everything is setup)

PR #11, opened 2024-02-16
Rewrite debug info: variable locations, integer types, formatting

cadmic:
Major changes:
* Add local variable register/stack locations
* Be explicit about which entries were "erased" (reconstructed via forensics, useful for functions that were inlined away)
* Remove erased global variables as they don't seem to be useful
* Use `s32` and friends (except `char`)
* Put offset comments in front of struct members and size comment to end of struct (matches OOT repo style)
* clang-format

A new file `debug/Fire/_buildtev.c` also appeared, I'm not sure why it wasn't in the original dump.

This was done using a modified version of dtk's dwarf dump tool (https://github.com/cadmic/decomp-toolkit/tree/oot-gc-debug-info) and then running clang-format on the output.

### DIFF
--- a/debug/Core/xlFileGCN.c
+++ b/debug/Core/xlFileGCN.c
@@ -1,156 +1,214 @@
-﻿// Location: 0x0
-long float _half$localstatic0$sqrtf__Ff;
+/*
+    Compile unit: C:\HOMEBOY\STEPHEN\Japanese Ocarina\Core\xlFileGCN.c
+    Producer: MW EABI PPC C-Compiler
+    Language: C++
+    Code range: 0x80005E68 -> 0x80006280
+*/
 
-// Location: 0x0
-long float _three$localstatic1$sqrtf__Ff;
+#include "types.h"
 
-// Location: 0x0
-long float _half$localstatic0$sqrt__Ff;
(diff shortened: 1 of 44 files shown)


debug/Core/xlFileGCN.c
```diff
@@ -1,156 +1,214 @@
...
+/*
+    Compile unit: C:\HOMEBOY\STEPHEN\Japanese Ocarina\Core\xlFileGCN.c
+    Producer: MW EABI PPC C-Compiler
+    Language: C++
+    Code range: 0x80005E68 -> 0x80006280
+*/
+
+#include "types.h"
+
+struct _XL_OBJECTTYPE {
```
Yanis42, line 10: it would be perfect if the structs are typedeffed, thanks a lot for this, it's gonna save a lot of time

cadmic: Good idea, I've given them all anonymous typedefs, but we still use `struct _CPU*` or `struct __anon_0xABCD*` when they're referenced from elsewhere in the debug info which I think is fine

Thar0 (approved):
Looks much better afaict, nice work

Yanis42 (approved):

PR #12, opened 2024-02-17
Match every functions from ``ram.s``

Yanis42:
this was fun, ``ramEvent`` was easy to match, I literally copy pasted ``romEvent`` and that's it lol (I also changed the first argument of this one since the callback is a void technically)

thanks to the people who helped me on the tiny issues I encountered


src/ram.c
```diff
@@ -0,0 +1,351 @@
...
+static s32 ramPutControl64(Ram* pRAM, u32 nAddress, s64* pData);
+static s32 ramPutControl16(Ram* pRAM, u32 nAddress, s32* pData);
+static s32 ramPutControl32(Ram* pRAM, u32 nAddress, s16* pData);
+static s32 ramPutControl8(Ram* pRAM, u32 nAddress, s8* pData);
+
+_XL_OBJECTTYPE gClassRAM = {
+    "RAM",
+    sizeof(Ram),
+    NULL,
+    (EventFunc)ramEvent,
```
cadmic, line 40: is this (EventFunc) cast necessary any more? ditto in rom.c

include/ram.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _RAM_H
+#define _RAM_H
+
+#include "types.h"
+
+typedef struct Ram {
```
cadmic, line 6: could you comment the `// __anon_name` above here for easy searching from debug info?

include/ram.h
```diff
@@ -0,0 +1,12 @@
...
+    /* 0x00 */ void* pHost;
+    /* 0x04 */ void* pBuffer;
+    /* 0x08 */ unsigned int nSize;
```
cadmic, line 9:
```suggestion
    /* 0x08 */ u32 nSize;
```

src/ram.c
```diff
@@ -0,0 +1,351 @@
...
+
+s32 ramWipe(Ram* pRAM) {
+    if (pRAM->nSize != 0 && !xlHeapFill32(pRAM->pBuffer, pRAM->nSize, 0)) {
+        return 0;
+    }
+
+    return 1;
+}
+
+// int ramGetBuffer(Ram *pRAM, void *ppRAM, unsigned int nOffset, unsigned int *pnSize);
```
cadmic, line 142: remove comment?

src/ram.c
```diff
@@ -0,0 +1,351 @@
...
+s32 ramGetBuffer(Ram* pRAM, void* ppRAM, u32 nOffset, u32* pnSize) {
```
cadmic, line 143: `ppRAM` makes it sound like the type is actually `void**`, I don't think the debug info can tell the difference it seems. I think there's evidence in the code too, it's assigning a pointer to `*ppRAM`, so `ppRAM` is a pointer to a pointer

src/ram.c
cadmic: actually the debug info doesn't have `romEvent` or `ramEvent` as taking `void*`. Does it not match if `ramEvent` takes a `Ram*`?

Yanis42: I changed it too in system

src/ram.c
Yanis42: oops lol

src/ram.c
Yanis42: debug infos says ``int ramGetBuffer(__anon_0x4BFE7 *pRAM, void *ppRAM, unsigned int nOffset, unsigned int *pnSize);`` 🤷 

though I agree that this name's weird, I just wanted to stay consistent

cadmic: I know, I'm saying that the debug info has it wrong in this case, I've noticed this somewhere in rom.c too. Probably the compiler doesn't distinguish between void** and void*, because in DWARF 1 `void*` is considered a "fundamental type" rather than being a pointer type

src/ram.c
Yanis42: it doesn't work without the void type because that's what the callback typedef expects, so it's either cast or use void* (I prefer using void as it is what oot-vc does)

cadmic: Right, but based on the debug info I think it's more likely that it was originally

```
int ramEvent(Ram* pRAM, int nEvent, void* pArgument);
```

and cast it as `(EventFunc)ramEvent` when creating this `XL_OBJECTTYPE` value. oot-vc didn't have that to go off of, we have to cast in one place or another (either in the function body, or here) and I think oot-vc guessed wrong

Yanis42: I'll revert this then

cadmic (approved):

PR #14, opened 2024-02-17
Import CPU types

cadmic:
I really hate the naming scheme (both `_CPU_ADDRESS` and `cpu_function`) so I gave everything CamelCase names.

In addition, I fixed the signatures of `cpuSetDevicePut` and `cpuSetDeviceGet` and removed `UnknownDeviceStruct`. It seems like for all these "objects" (ROM, RAM, etc.) have a `void* pHost` which is actually a `System`, and when receiving event 0x1002 the object should register its get/set functions with a `CpuDevice` based on the device type.

### DIFF
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -3,17 +3,159 @@
 
 #include "types.h"
 
-typedef s32 (*Put8Func)(struct Rom* pROM, u32 nAddress, s8* pData);
-typedef s32 (*Put16Func)(struct Rom* pROM, u32 nAddress, s16* pData);
-typedef s32 (*Put32Func)(struct Rom* pROM, u32 nAddress, s32* pData);
-typedef s32 (*Put64Func)(struct Rom* pROM, u32 nAddress, s64* pData);
+//! TODO: Move to Dolphin SDK headers
+typedef struct OSContext {
+    /* 0x000 */ u32 gpr[32];
+    /* 0x080 */ u32 cr;
+    /* 0x084 */ u32 lr;
+    /* 0x088 */ u32 ctr;
+    /* 0x08C */ u32 xer;
+    /* 0x090 */ f64 fpr[32];
+    /* 0x190 */ u32 fpscr_pad;
+    /* 0x194 */ u32 fpscr;
+    /* 0x198 */ u32 srr0;
+    /* 0x19C */ u32 srr1;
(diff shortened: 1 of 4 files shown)


include/cpu.h
```diff
@@ -27,12 +169,106 @@ typedef struct __anon_0x3EB4F {
...
+    /* 0x18 */ u32 addr_check;
+    /* 0x1C */ s32 addr_last;
+    /* 0x20 */ u32 checkType;
+    /* 0x24 */ u32 checkNext;
+} CpuOptimize; // size = 0x28
+
+typedef struct Cpu Cpu;
+
+// _CPU
+struct Cpu {
```
Yanis42, line 222: is it just how it was originally? it looks weird to me

src/rom.c
```diff
@@ -158,24 +158,26 @@ s32 romEvent(Rom* pROM, s32 nEvent, void* pArgument) {
             }
             break;
         case 0x1002:
-            switch (((UnknownDeviceStruct*)pArgument)->unk) {
+            switch (((CpuDevice*)pArgument)->nType) {
                 case 0:
```
Yanis42, line 162: is there an enum for events/cpu device type?

Yanis42 (commented):
nice work!

src/rom.c
cadmic: not yet, at least here's no enum with value 0x1002 or 4098 in the debug info. Maybe something fancy is going on with bitfields? (like 0x1000 means something and so does 0x2)

include/cpu.h
cadmic: Do you mean the separate typedef? I don't think we can know how it was originally, but if you want to have a C struct refer to itself, doing

```
typedef struct MyStruct {
  MyStruct* data;
} MyStruct;
```
doesn't work, instead you need to do something like
```
typedef struct MyStruct {
  struct MyStruct* data;
} MyStruct;
```
or
```
typedef struct MyStruct MyStruct;

struct MyStruct {
 MyStruct* data;
};
```
It's a bit confusing because there's both `struct MyStruct` (the original struct) and `MyStruct` (the typedef) floating around

Yanis42: oooh yea right, I forgot some of the structs are recursive like this

Yanis42 (approved):

PR #15, opened 2024-02-17
Set -inline deferred and reverse function order

cadmic:
I'm fairly sure `-inline deferrred` was used, because I don't see how the inlining can work out for `romCacheGame` otherwise. This has the side-effect of reversing the function order between the source code and the compiled object, which makes a lot of sense logically and it's consistent with the line numbers in the debug info.

I've also updated func_splitter.py to split the functions in reverse order now. Probably we should try to update the debug info too, but I'll do that in another PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ LDSCRIPT := $(BUILD_DIR)/ldscript.lcf
 
 # Outputs
 ELF := $(BUILD_DIR)/$(TARGET).elf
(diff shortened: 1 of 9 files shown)


Yanis42 (approved):
I think you are correct, is you look at ram.c on oot-vc you'll see that ramEvent is the last function, on current main it's at the top (and I assume both files are in the opposite order)

nice find!

PR #17, opened 2024-02-17
Match sram.c

louist103:


src/sram.c
```diff
@@ -0,0 +1,132 @@
+#include "types.h"
```
cadmic, line 1: could you rebase to pick up the `-inline deferred` change and reverse the function order? Sorry for the disruption

asm/non_matchings/sram/sramCopySRAM.s
```diff
@@ -0,0 +1,34 @@
+glabel sramCopySRAM
```
cadmic, line 1: could we delete these files now? We can keep the original asm around for reference

src/sram.c
```diff
@@ -0,0 +1,132 @@
...
+static int sramPut64(void* pObject, u32 nAddress, s64* pData);
+static int sramPut32(void* pObject, u32 nAddress, s32* pData);
+static int sramPut16(void* pObject, u32 nAddress, s16* pData);
+static int sramPut8(void* pObject, u32 nAddress, s8* pData);
+
+extern int ramGetBuffer(struct Ram *pRAM, void *ppRAM, u32 nOffset, u32 *pnSize);
+extern int simulatorWriteSRAM(unsigned int address, unsigned char *data, int size);
+
+_XL_OBJECTTYPE gClassSram = {
+    OBJ_NAME, sizeof(OBJ_NAME) - 1, NULL, sramEvent
```
cadmic: should be `sizeof(Sram)` not `sizeof(OBJ_NAME)` (probably don't need an `OBJ_NAME` macro now either)

Also, is it not necessary to do `(EventFunc)sramEvent` here to satisfy the compiler?

src/sram.c
```diff
@@ -0,0 +1,132 @@
...
+typedef struct Sram
+{
+	struct sramHost *pHost; // 0x0
+} Sram;
+// __anon_0x74AB9
+
+struct Cpu;
+struct Ram;
+
+typedef struct sramHost{
```
cadmic: Sram should have `void* pHost` but it's secretly a `System*`, and we've been accessing it like this:

```
Cpu* pCPU = ((System*)obj->pHost)->apObject[SOT_CPU]
Ram* pRAM = ((System*)obj->pHost)->apObject[SOT_RAM]
```
See e.g. `romEvent` or `ramEvent` in the main branch

Probably we should make macros for this eventually.

src/sram.c
```diff
@@ -0,0 +1,132 @@
...
+int simulatorWriteSRAM(unsigned int address, unsigned char *data, int size);
+int simulatorReadSRAM(unsigned int address, unsigned char *data, int size);
+
+#define OBJ_NAME "SRAM"
+
+typedef struct Sram
+{
+	struct sramHost *pHost; // 0x0
+} Sram;
+// __anon_0x74AB9
```
cadmic, line 12: nit: please match the style of e.g. ram.h here (with offset comments in front for example). `make format` can help

src/sram.c
```diff
@@ -0,0 +1,132 @@
...
+    switch (nEvent) {
+        case 2:
+            pSram->pHost = pArgument;
+        case 0:
+        case 1:
+        case 3:
+            break;
+
+        case 0x1002:
+            if (cpuSetDevicePut(pSram->pHost->pCpu, pArgument, sramPut8, sramPut16, sramPut32, sramPut64) == 0) {
```
cadmic, line 53: nit: `func() == 0` -> `!func()` (I imagine we'll have a `bool` type eventually)

src/sram.c
```diff
@@ -0,0 +1,132 @@
...
+
+        default:
+            return 0;
+    }
+    return 1;
+}
+
+// TODO uintptr_t?
+static int sramGet64(void* pObject, u32 nAddress, s64* pData) {
+    simulatorReadSRAM(nAddress & 0x7FFF, (unsigned char*)pData, sizeof(long long));
```
cadmic, line 70: `u8`, `s64` etc

src/sram.c
```diff
@@ -0,0 +1,117 @@
...
+    if (!ramGetBuffer(((System*)pSRAM->pHost)->apObject[SOT_RAM], &pTarget, nOffsetRAM, &nSize)) {
+        return 0;
+    }
+    if (!simulatorWriteSRAM(nOffsetSRAM & 0x7FFF, pTarget, nSize)) {
+        return 0;
+    }
+    return 1;
+}
+
+static int sramPut8(void* pObject, u32 nAddress, s8* pData) {
```
cadmic, line 50: similarly, these should all be `sramPut8(Sram* pSram, ...)` with a cast to `(Put8Func)` etc where it's used. This isn't in the debug info but the pattern is pretty clear from other devices like rom/ram etc.

src/sram.c
```diff
@@ -0,0 +1,117 @@
+#include "types.h"
+#include "xlObject.h"
+#include "cpu.h"
+#include "system.h"
+#include "xlObject.h"
```
cadmic, line 5: nit: "xlObject.h" included twice

cadmic (approved):

PR #18, opened 2024-02-17
Match ``movie.c``, ``codeGCN.c``, ``disk.c`` and ``serial.c``

Yanis42:
they're all the same functions, except movie.c, also some structs were missing from debug infos but based on how this class system is designed I made some that are consistent (like ``Disk``)

### DIFF
--- a/include/disk.h
+++ b/include/disk.h
@@ -1,0 +1,13 @@
+#ifndef _DISK_H
+#define _DISK_H
+
+#include "types.h"
+
+// based on how other "gClass" works, we can assume this was the struct used for Disk
+typedef struct Disk {
+    /* 0x00 */ void* pHost;
+} Disk; // size = 0x04
+
+s32 diskEvent(Disk* pDisk, s32 nEvent, void* pArgument);
+
+#endif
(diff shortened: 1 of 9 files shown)


src/movie.c
```diff
@@ -0,0 +1,70 @@
...
+    /* 0x18 */ unsigned char field_rendering;
+    /* 0x19 */ unsigned char aa;
+    /* 0x1A */ unsigned char sample_pattern[12][2];
+    /* 0x32 */ unsigned char vfilter[7];
+} GXRenderModeObj; // size = 0x3C
+
+extern GXRenderModeObj* rmode;
+extern int __OSCurrHeap;
+
+void* gBufferP;
```
cadmic, line 43: nit: I'd swap this with the `OSAllocFromHeap` declaration so it doesn't look like part of the SDK

cadmic (approved):

include/disk.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _DISK_H
+#define _DISK_H
+
+#include "types.h"
+
+// based on how other "gClass" works, we can assume this was the struct used for Disk
```
cadmic, line 6:
```suggestion
// __anon_0x73B29
```
turns out this one is in the debug info after all

cadmic (approved):

PR #19, opened 2024-02-17
Match the rest of rom.c

cadmic:
`romCacheGame` requires some auto-inlining setting (guessing `-inline auto` for now), which breaks xlList so I had to un-match a function there.

Thanks to @Yanis42 for doing most of `romLoadUpdate` and `romCopyUpdate`


Yanis42 (approved):
nice!!

PR #20, opened 2024-02-17
Reverse function order in debug info

cadmic:
This should hopefully make it easier to compare with the source code, since the order in the compiled objects is reversed due to `-inline deferred`


Yanis42 (approved):
there's not a lot to review here as it's just debug informations, I agree with the change as it makes it consistent with the rest of the codebase

PR #22, opened 2024-02-18
Import SDK headers from dolsdk2001

cadmic:
It seems like every GC/Wii decomp project has their own set of dolphin SDK headers. I propose we standardize on [dolsdk2001](https://github.com/doldecomp/dolsdk2001), both because it's seems cleaner to use an SDK-related project as a "source of truth" rather than some random game, and it will be easier to keep our headers in sync.

I got rid of types.h in favor of <dolphin/types.h>, since it probably doesn't make sense to redefine everything there.


Yanis42:
Is it fine to format these files? I think it's weird to have some inconsistency here, I know they're SDK files but I think we should keep formatting consistent

Yanis42:
also I don't really understand why using ``<>`` for the includes, isn't this supposed to be used when including headers outside the project? like ``#include <string.h>``

cadmic:
> Is it fine to format these files?

This might make it harder to sync later if we make local changes but it's probably fine, so I've formatted them.

> also I don't really understand why using `<>` for the includes, isn't this supposed to be used when including headers outside the project? like `#include <string.h>`

This is what the dolsdk2001 headers use so I thought it was more consistent with that. It is importing from a "standard library" of sorts rather than from the project (the GC emu itself) so it kinda makes sense but idk

Yanis42 (approved):

Thar0 (approved):
I'm not sure about using `<>`-style includes for anything. It can cause some conflicts with other standard includes, but this is probably only a problem very long-term (say if someone tried to recompile the whole emulator with a newer compiler) and should be fairly easy to fix in the future

PR #25, opened 2024-02-21
Match most of system.c

Yanis42:
This took a long time but here it is! shoutouts to everyone who helped on this
also I was able to remove all of the static data at the top of the file (except few things of course)


include/system.h
```diff
@@ -83,6 +131,13 @@ typedef struct SystemRomConfig {
     /* 0x0170 */ s32 currentControllerConfig;
 } SystemRomConfig; // size = 0x174
 
+#define SYSTEM_CPU(pSystem) ((struct Cpu*)(((System*)(pSystem))->apObject[SOT_CPU]))
```
cadmic, line 134: hmm should these go in macros.h?

Also no need for `struct Cpu` instead of `Cpu`

src/movie.c
```diff
@@ -2,17 +2,20 @@
 #include <dolphin/os.h>
 #include <dolphin/types.h>
 #include <dolphin/vi.h>
+#include "system.h"
 
 extern GXRenderModeObj* rmode;
 void* gBufferP;
 
+s32 simulatorGetArgument(SystemArgumentType eType, char** pszArgument);
```
cadmic, line 10: shouldn't this go in simGCN.h?

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
 #include "xlObject.h"
 #include "system.h"
+#include "rom.h"
+#include "ram.h"
+#include "cpu.h"
+#include "mcard.h"
+#include "pif.h"
+#include "simGCN.h"
+
+extern _XL_OBJECTTYPE gClassRdb;
```
cadmic, line 11:
```suggestion
//! TODO: move these declarations to the proper headers
extern _XL_OBJECTTYPE gClassRdb;
```

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
+static s32 systemGet8(System* pSystem, u32 nAddress, s8* pData);
+static s32 systemGet16(System* pSystem, u32 nAddress, s16* pData);
+static s32 systemGet32(System* pSystem, u32 nAddress, s32* pData);
+static s32 systemGet64(System* pSystem, u32 nAddress, s64* pData);
+static s32 systemPut8(System* pSystem, u32 nAddress, s8* pData);
+static s32 systemPut16(System* pSystem, u32 nAddress, s16* pData);
+static s32 systemPut32(System* pSystem, u32 nAddress, s32* pData);
+static s32 systemPut64(System* pSystem, u32 nAddress, s64* pData);
+
+s32 mipsResetInterrupt(Mips* pMips, MipsInterruptType eType);
```
cadmic, line 46: these function declarations could probably move now though

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
-char D_800EB60C[] = "PokemonStadium";
-char D_800EB61C[] = "Pokemon Stadium";
-char D_800EB62C[] = "POKEMONSTADIUM";
-char D_800EB63C[] = "Panel de Pon";
-char D_800EB64C[] = "YoshiStory";
-
-SystemRomConfig gSystemRomConfigurationList;
+}; // size = 0x10
+
+// clang-format off
```
cadmic, line 67: instead I think it would be better to group them explicitly:
```
static u32 contMap[4][20] = {
  {
    ...
  },
  {
    ...
  },
  {
    ...
  },
  {
    ...
  },
};

File: src/system.c [hunk 6]
```diff
@@ -1,167 +1,1219 @@
...
+
+        if (nSizeExtra > (s32)(nSizeCacheROM + 0xFFF00000)) {
+            nSizeExtra = nSizeCacheROM + 0xFFF00000;
+        }
+
+        nSizeRAM += nSizeExtra;
+        nSizeCacheROM -= nSizeExtra;
+    }
+
+    if (!ramSetSize(pSystem->apObject[SOT_RAM], nSizeRAM)) {
```
cadmic, line 171: can this be `SYSTEM_RAM`?

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
+        nSizeCacheROM = 0x400000;
+    } else {
+        nSizeRAM = 0x400000;
+        nSizeCacheROM = 0x800000;
+    }
+
+    if (simulatorGetArgument(SAT_XTRA, &szExtra)) {
+        nSizeExtra = atoi((char*)szExtra) << 0x14;
+
+        if (nSizeExtra > (s32)(nSizeCacheROM + 0xFFF00000)) {
```
cadmic, line 163:
```suggestion
        if (nSizeExtra > nSizeCacheROM + 0xFFF00000) {
```

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
+    if (bExpansion) {
+        nSizeRAM = 0x800000;
+        nSizeCacheROM = 0x400000;
+    } else {
+        nSizeRAM = 0x400000;
+        nSizeCacheROM = 0x800000;
+    }
+
+    if (simulatorGetArgument(SAT_XTRA, &szExtra)) {
+        nSizeExtra = atoi((char*)szExtra) << 0x14;
```
cadmic, line 161:
```suggestion
        nSizeExtra = atoi(szExtra) << 0x14;
```

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
+    if (romTestCode(pROM, "NSME") || romTestCode(pROM, "NSMJ")) {
+        // Super Mario 64
+        systemSetControllerConfiguration(&gSystemRomConfigurationList[index], 0x01010101, 0);
+        gSystemRomConfigurationList[index].storageDevice = SOT_RSP;
+
+    } else if (romTestCode(pROM, "CZLE") || romTestCode(pROM, "CZLJ")) {
+        // Ocarina of Time
+        gSystemRomConfigurationList[index].storageDevice = SOT_PIF;
+
+        if (!simulatorGetArgument(SAT_VIBRATION, &szText) || (*szText == 49)) {
```
cadmic, line 223: I think `'0'` and `'1'` would be better than `48` and `49` for these

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
+            if (!simulatorGetArgument(SAT_CONTROLLER, &szText) || (*szText == 48)) {
+                systemSetControllerConfiguration(&gSystemRomConfigurationList[index], 0x82828282, 1);
+            } else {
+                systemSetControllerConfiguration(&gSystemRomConfigurationList[index], 0x80808080, 1);
+            }
+        } else if (!simulatorGetArgument(SAT_CONTROLLER, &szText) || (*szText == 48)) {
```
cadmic, line 229: IMO this would be slightly cleaner to split the `else if` as
```
if () {
  if () {
  } else {
  }
} else {
  if () {
  } else {
  }
}
```
just to show off the parallel structure (ditto for MM)

cadmic (commented):
Great work peepoHappy

src/system.c
```diff
@@ -179,24 +1228,507 @@ const f64 D_80135F98 = 4503599627370496.0;
...
+                        }
+                        if (!cpuMapObject(pCPU, pSystem->apObject[SOT_VIDEO], 0x04400000, 0x044FFFFF, 0)) {
+                            return 0;
+                        }
+                        break;
+                    case SOT_SERIAL:
+                        if (!xlObjectMake(&pSystem->apObject[SOT_SERIAL], pSystem, &gClassSerial)) {
+                            return 0;
+                        }
+                        if (!cpuMapObject(pCPU, pSystem->apObject[SOT_SERIAL], 0x04800000, 0x04900000 - 1, 0)) {
```
cadmic, line 1651: 0x048FFFFF here and below to match the pattern above?

src/system.c
Yanis42: seems to work

src/system.c
Yanis42: it doesn't match without the cast apparently

src/system.c
```diff
@@ -1,167 +1,1219 @@
...
+    } else if (!pifSetEEPROMType(pPIF, CT_NONE)) {
+        return 0;
+    }
+
+    if (romTestCode(pROM, "NSME") || romTestCode(pROM, "NSMJ")) {
+        // Super Mario 64
+        pSystem->eTypeROM = SRT_MARIO;
+        nSizeSound = 0x1000;
+
+        if (DVDOpen(Z_ICON_PATH, &fileInfo) == 1 &&
```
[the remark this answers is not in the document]
Yanis42: yeah they are mandatory😔

cadmic (approved):

PR #28, opened 2024-02-21
Match xlPostGCN

Yanis42:
~~...except xlPostText, I have no idea how to match this one, I'm really lost on this one~~


asm/xlCoreGCN.s
```diff
@@ -625,6 +625,10 @@ D_800DB7BC:
 rmodeobj:
     .skip 60
 
+/* 000F0F9C 800F3F1C 0080 */
+glabel g_texMap
+    .skip 128
+
```
cadmic, line 631: shouldn't this be removed from somewhere else as well? It would be good to check that this still builds OK if all of the `src` are replaced by `asm` in obj_files.mk

Yanis42: should be good now

cadmic (approved):

PR #29, opened 2024-02-21
Start ``simGCN.c``

Yanis42:
thanks to cadmic for figuring out the data stuff


include/simGCN.h
```diff
@@ -49,11 +49,61 @@ typedef struct __anon_0xDB69 {
...
+    S_M_CARD_SV11 = 44,
+    S_M_CARD_SV12 = 45,
+    S_M_CARD_SV_SHARE = 46,
+    S_M_CARD_DEFAULT_ERROR = 47,
+} __anon_0x61D7;
+
 extern s32 gDVDResetToggle;
 extern s32 gbDisplayedError;
 
-extern System* gpSystem;
```
cadmic, line 55: why remove this declaration?

src/simGCN.c
```diff
@@ -0,0 +1,346 @@
+#include "simGCN.h"
+
+// TODO: move to macros.h
+#define ALIGNAS(N) __attribute__((aligned(N)))
```
cadmic, line 4: sorry for making you clean up my mess, but could you move this now? It seems like will probably be useful for other files soon

src/simGCN.c
```diff
@@ -0,0 +1,346 @@
...
+
+s32 gHighlightChoice = 0x00000001;
+__anon_0x61D7 simulatorMessageCurrent = S_M_NONE;
+s32 gResetBeginFlag = 0x00000001;
+
+char D_80134D9C[8] = "yes.raw";
+char D_80134DA4[7] = "no.raw";
+
+// TODO: struct
+static s32 gpCode;
```
cadmic, line 227: should we make this struct now? (__anon_0x57A1)

include/simGCN.h
Yanis42: because the debug data says it's declared in simGCN.c so it doesn't make sense to extern it in this header

Yanis42: maybe I misread

src/simGCN.c
Yanis42: I named it "Code", we can rename it later if it turns out it's a bad name

cadmic: Yeah I think that is too generic (I would have left it anonymous myself) but I was mostly concerned about getting the struct layout right when matching functions

PR #30, opened 2024-02-21
Kinda fix decompme.py

cadmic:
We now run the actual preprocessor so at least the context will compile correctly. The downside is that macros are not preserved in the context, and decomp.me's automatic m2c decompilation doesn't work.


cadmic:
See #46 instead

PR #32, opened 2024-02-21
Tweak headers and use SYSTEM_ getters everywhere

cadmic:
Basically I used them whenever the types made sense.

I changed my mind and moved them back to system.h because it seems more related to "system" than as a general "macro". In the process I added some missing headers.


Yanis42 (approved):
nice!

PR #33, opened 2024-02-22
Match xlMain

cadmic:
and simulatorGetArgument, and shuffle some stuff around


include/simGCN.h
```diff
@@ -104,11 +113,14 @@ extern void* gpSound;
 extern System* gpSystem;
 
 void simulatorUnpackTexPalette(__anon_0xDB69* pal);
+s32 simulatorDVDOpen(char* szNameFile, struct DVDFileInfo* pFileInfo);
+s32 simulatorDVDRead(struct DVDFileInfo* pFileInfo, void* anData, s32 nSizeRead, s32 nOffset,
```
Yanis42: is ``struct`` necessary here? should be a sdk thing right?

include/simGCN.h
```diff
@@ -104,11 +113,14 @@ extern void* gpSound;
...
+                     void (*callback)(s32, struct DVDFileInfo*));
```
Yanis42: the callback here should be ``DVDCallback`` from dvd.h btw

src/simGCN.c
```diff
@@ -343,6 +355,222 @@ const f32 D_80135D6C = 120.0;
...
+s32 xlMain() {
+    // Local variables
+    GXColor color; // r1+0x3C
+    SystemMode eMode; // r1+0x38
+    s32 nSize0; // r1+0x34
+    s32 nSize1; // r1+0x30
+    s32 iName; // r5
+    char* szNameROM; // r1+0x2C
+    char acNameROM[32]; // r1+0xC
+    // s32 rumbleYes; // r1+0x8
```
Yanis42, line 389: should we keep these comments? I usually remove them once I'm done matching a function

Yanis42 (commented):
small things

include/simGCN.h
Yanis42: nvm you changed that before I sent the review lol

src/simGCN.c
cadmic: I was too excited and forgot to clean up

cadmic: I kept `// s32 rumbleYes;` because it's in the debug data and I'm not sure what to do with it, but someone else might

Yanis42 (approved):

PR #34, opened 2024-02-22
Match some simGCN functions

Yanis42:
some functions match but are still ifdeffed because of data issues (which should be solved once the remaining functions are matched)

---
<details>
Match:

- simulatorGXInit
- simulatorDVDOpen
- simulatorDVDRead
- simulatorPlayMovie
- simulatorReset
- simulatorSetControllerMap
- simulatorCopyControllerMap
- simulatorShowLoad
- simulatorDetectController
- simulatorReadPak
- simulatorWritePak
- simulatorReadEEPROM
- simulatorWriteEEPROM
- simulatorReadSRAM
- simulatorWriteSRAM
- simulatorReadFLASH
- simulatorWriteFLASH
- simulatorRumbleStart
- simulatorRumbleStop
- simulatorTestReset
- simulatorResetAndPlayMovie
</details>


src/rom.c
```diff
@@ -1106,7 +1106,7 @@ s32 romUpdate(Rom* pROM) {
 
         nStatus = DVDGetCommandBlockStatus(&pROM->fileInfo.cb);
         if (nStatus != 1) {
-            if (!simulatorDVDShowError(nStatus, pROM->load.anData, pROM->load.nSizeRead,
+            if (!simulatorDVDShowError(nStatus, (s32)pROM->load.anData, pROM->load.nSizeRead,
```
cadmic, line 1109: can this cast go away if `simulatorDVDShowError` takes `u32` as the second argument? (Maybe the third argument too based on this call site)

src/simGCN.c
```diff
@@ -281,17 +305,189 @@ const f32 D_80135D64 = 10000.0;
...
+        do {
+            bRetry = 0;
+            DVDReadAsyncPrio(pFileInfo, anData, nSizeRead, nOffset, NULL, 2);
+
+            while ((nStatus = DVDGetCommandBlockStatus(&pFileInfo->cb)) != 0) {
+                if (!simulatorDVDShowError(nStatus, (s32)anData, nSizeRead, nOffset)) {
+                    return 0;
+                }
+
+                if ((nStatus == 0xB) || (nStatus == -1)) {
```
cadmic, line 472:
```suggestion
                if (nStatus == 11 || nStatus == -1) {
```
11 instead of 0xB because this is probably an enum

and personally I really hate that the oot codebase uses redundant parentheses so much, personally I think it makes things harder to read. There's more redundant parentheses below but I haven't really tried to enforce this so don't worry about it

src/simGCN.c
```diff
@@ -309,41 +505,283 @@ const f32 D_80135D6C = 120.0;
...
+    simulatorResetInit();
 
-#pragma GLOBAL_ASM("asm/non_matchings/simGCN/simulatorSetControllerMap.s")
+    if (IPL == 1) {
+        if (forceMenu == 1) {
+            OSResetSystem(1, 0, 1);
+            return;
+        }
+        OSResetSystem(1, 0, 0);
+        return;
```
cadmic, line 528: can this be
```
if (forceMenu == 1) {
    OSResetSystem(1, 0, 1);
} else {
    OSResetSystem(1, 0, 0);
}
return;
```
?

src/simGCN.c
```diff
@@ -309,41 +505,283 @@ const f32 D_80135D6C = 120.0;
...
+        DemoCurrentBuffer = DemoFrameBuffer2;
+    } else {
+        DemoCurrentBuffer = DemoFrameBuffer1;
+    }
+}
+
+void simulatorResetAndPlayMovie(void) {
+    int pad1;
+    GXColor color; // r1+0x14
+    GXRenderModeObj* simrmode; // r31
```
cadmic:
```suggestion
    GXColor color;
    GXRenderModeObj* simrmode;
```

src/simGCN.c
```diff
@@ -309,41 +505,283 @@ const f32 D_80135D6C = 120.0;
...
+        pifWriteRumble(SYSTEM_PIF(gpSystem), channel, address, data);
+    }
+
+    return 1;
+}
 
+s32 simulatorReadEEPROM(u8 address, u8* data) {
+    s32 size;
+
+    if (pifGetEEPROMSize(gpSystem->apObject[1], &size, gpSystem) == 0) {
```
cadmic, line 658: SYSTEM_PIF?

Also, does it have be`== 0` instead of `!`?

src/simGCN.c
```diff
@@ -309,41 +505,283 @@ const f32 D_80135D6C = 120.0;
...
+    return 1;
+}
+
+// function match but data don't
+#ifndef NON_MATCHING
 #pragma GLOBAL_ASM("asm/non_matchings/simGCN/simulatorTestReset.s")
+#else
+inline s32 simulatorTestResetUnknownInline(u32 nTick) {
+    s32 result = (f32)(nTick - gnTickReset) >= (0.5f * (f32)(*(u32*)0x800000F8 >> 2));
+    return result;
```
cadmic, line 713: should be `s32 result = (nTick - gnTickReset >= OSSecondsToTicks(0.5f));`

also are you sure this is an inline function? This could conceivably be written directly

src/rom.c
Yanis42: 😔 
```
src\rom.c:1110: cannot convert
src\rom.c:1110: 'unsigned char *' to
src\rom.c:1110: 'long'
```

Yanis42: oh I wonder if it's because the other parameters aren't named

cadmic: Oh, what if we change it to u8*?

Yanis42: yeah ok doesn't work, probably because it's a pointer

Yanis42: I still did the signed -> unsigned change

src/simGCN.c
Yanis42: I usually put everything in parenthesis because sometimes the condition doesn't behave like I want

cadmic (approved):

PR #38, opened 2024-02-24
Rework make setup

Thar0:
- Remove dependence on devkitpro, instead downloads a suitable binutils release from https://github.com/encounter/gc-wii-binutils
- Download compilers on setup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ include obj_files.mk
 # Tools
 #-------------------------------------------------------------------------------
 
-MWCC_VERSION := 1.1
+MWCC_VERSION := GC/1.1
 
 # Programs
 ifeq ($(WINDOWS),1)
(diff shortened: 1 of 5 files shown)


Yanis42 (approved):
nice!

tools/.gitignore
```diff
@@ -0,0 +1,2 @@
+binutils/**
+mwcc_compiler/**
```
cadmic, line 2: Why a separate .gitignore? I don't particularly care but there's already some `tools/` stuff gitignored in the root dir so it seems a bit messier now

Makefile
```diff
@@ -42,7 +42,7 @@ else
 	WINE := wine
 endif
 
-PPC_BIN_PREFIX := $(DEVKITPPC)/bin/powerpc-eabi-
+PPC_BIN_PREFIX := tools/binutils/powerpc-eabi-
```
cadmic, line 45: Could you quickly update OBJDUMP_EXECUTABLE in objdump_wrapper.py? Currently it assumes that powerpc-eabi-objdump is in your PATH but now we don't have to.

cadmic (commented):
Works on my machine (x86 mac)

cadmic (approved):

PR #39, opened 2024-02-24
Match cpuCompile functions

cadmic:
All of these allocate a buffer and write some PowerPC instructions to it to simulate a MIPS instruction. A couple are inlined into cpuExecute but I did them now too. I added comments with some disassembly using [a script](https://gist.github.com/cadmic/75664feff2bfabf530ff6a439e003a5f).


src/cpu.c
```diff
@@ -524,55 +524,1118 @@ const f32 D_80135FB8 = 0.5;
...
+    compile[count++] = 0x4E800020; // blr
+
+    nSize = count;
+    DCStoreRange(compile, nSize * sizeof(s32));
+    ICInvalidateRange(compile, nSize * sizeof(s32));
+
+    return 1;
+}
+
+static s32 cpuCompile_W_CVT_SD(Cpu* pCPU, s32* addressGCN) {
```
Yanis42, line 1094: the only thing I'd say about this pr would be adding comments describing what these names are, I guess it's something like "word something store double"?

I guess it makes sense if you know your way around mips but I think it would be nice to have that written somewhere (ok some of them are explicit but for this one example I'm kinda lost and there's others like this)

Yanis42 (commented):
nice work!

src/cpu.c
cadmic: to be honest I don't know what this one does, apparently there's `cvt.w.s` and `cvt.w.d` but I'm not sure if this somehow implements both of them or there's another instructions. Maybe it will be clearer once more of cpu is decomped?

Yanis42 (approved):

PR #40, opened 2024-02-24
Introduce PAD_STACK() and match xlList with it

cadmic:
```
inline void padStack(void) { s32 pad = 0; }
```

I needed this for cpuExecute to adjust the stack of inlined functions and local variables didn't seem to do anything. I just realized it works for xlList too.

I also turned `NO_INLINE` into `NO_INLINE()` since I think that looks a little better when it's "function-like".


Yanis42 (approved):
🆗

PR #45, opened 2024-02-25
Document ``contMap`` configurations

Yanis42:
This adds an enum to know which gamecube controller button it is in the array and there's macros for n64 buttons, which are used as the elements of the second array, some values are still in hex because I think they should be the left stick mappings but they don't do anything (at least it was the same behavior) when you map a button (note that they share the same mappings as the D-Pad)

I tried using clang's ``AlignTrailingComments`` option but it was doing strange things in functions.h so I decided to manually disable the formatting on the array for now (I think the comments looks more readable when aligned)

### DIFF
--- a/include/system.h
+++ b/include/system.h
@@ -24,7 +24,7 @@
 typedef s32 (*SystemCopyCallbackFunc)(void);
 
 // note: each stick direction count as an input
-typedef enum GcnButtons {
+typedef enum GcnButton {
     GCN_BTN_A = 0,
     GCN_BTN_B = 1,
     GCN_BTN_X = 2,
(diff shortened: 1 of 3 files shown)


include/system.h
```diff
@@ -5,8 +5,49 @@
...
+#define N64_BTN_CUP (1 << 19)
+#define N64_BTN_CDOWN (1 << 18)
+#define N64_BTN_CLEFT (1 << 17)
+#define N64_BTN_CRIGHT (1 << 16)
+#define N64_BTN_UNSET 0
+
 typedef s32 (*SystemCopyCallbackFunc)(void);
 
+// note: each stick direction count as an input
+typedef enum GcnButtons {
```
cadmic, line 27: nit: should be singular

cadmic (approved):

PR #46, opened 2024-02-25
Handle more preprocessor directives in decompme script

cadmic:
Also INCBIN and sync compiler flags. Unfortunately you still have to do a lot of manual fixing of the context (but at least you now don't have to delete all previous function definitions like in my first attempt).

Tested with a cpu.c function and a simGCN.c function, but there could be weirder stuff in other files

### DIFF
--- a/tools/decompme.py
+++ b/tools/decompme.py
@@ -21,16 +21,14 @@ COMPILER_NAME = "mwcc_233_159"
 # We don't set -inline deferred because otherwise the reversed function order
 # would require manually deleting all previous function definitions from the
 # context.
-COMPILER_FLAGS = (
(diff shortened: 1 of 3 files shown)


tools/decompme.py
```diff
@@ -65,6 +86,10 @@ def main():
     context = process_file(args.c_file)
     source_code = ""  # TODO: separate source code from context automatically
 
+    if args.print_context:
+        print(context)
+        return
+
```
Yanis42, line 92: the ``json_data: Dict[str, str] = json.loads(resp)`` line below (can't comment that line apparently) should be ``json_data: dict[str, str] = json.loads(resp)`` I think

Yanis42 (commented):
just a tiny thing otherwise seems ok

Yanis42 (approved):

PR #47, opened 2024-02-26
Remove compile.sh

cadmic:
compile.sh is pretty hard to debug. It's less mysterious if the commands were written in the Makefile instead

~~Also, encode Japanese strings via iconv so we can write them directly~~


cadmic:
Drafting because the iconv stuff doesn't work on ubuntu

cadmic:
I reverted the Japanese encoding, will revisit later

PR #49, opened 2024-02-26
Setup the remaining files from Fire and Core for decomp

Yanis42:
This builds 🆗 but there's dumb things, such as declaring globals as static, also sorry for the big pr, I did one commit per file so should be ok-ish to review


src/THPRead.c
```diff
@@ -0,0 +1,80 @@
...
+static OSMessageQueue ReadedBufferQueue;
+static OSMessageQueue ReadedBufferQueue2;
+static void* FreeReadBufferMessage[10];
+static void* ReadedBufferMessage[10];
+static void* ReadedBufferMessage2[10];
+static OSThread ReadThread;
+static u8 ReadThreadStack[4096];
+f32 gOrthoMtx[4][4];
+
+//! TODO: remove the suffix
```
cadmic, line 32: nit: maybe comment why the suffix is there or why it can't be removed right now (should be marked static but can't because of data ordering issues, and there's another global variable of the same name)

cadmic (approved):
Amazing

I didn't look too closely because we'll go into more detail when decomping the individual files but this makes things a whole lot easier. Thanks for improving the headers too

PR #51, opened 2024-02-28
Match ``xlFileGCN.c`` and ``xlObject.c`` (1 left)

Yanis42:
the only function left is ``xlObjectMake``, this is an annoying one because of all of the pointer operations, it's very likely that the node struct is fake but couldn't get something better without it (currently only regalloc left)

also the inline for ``xlObjectEvent`` doesn't match the erased function for the double pointer, not sure why


include/xlFileGCN.h
```diff
@@ -3,11 +3,15 @@
 
 #include "dolphin.h"
 
-typedef enum __anon_0x2757 {
+typedef s32 (*DVDOpenCallback)(char*, DVDFileInfo*);
+typedef s32 (*DVDReadCallback)(DVDFileInfo*, void*, s32, s32, DVDCallback);
+
+// XlFileType
```
cadmic, line 9:
```suggestion
// __anon_0x2757
```

src/xlFileGCN.c
```diff
@@ -0,0 +1,149 @@
...
+            return 0;
+        }
+
+        return 1;
+    }
+
+    return 0;
+}
+
+inline s32 xlFileGetFile(tXL_FILE** ppFile, char* szFileName) {
```
cadmic, line 44: does this inline work if it takes `tXL_FILE* pFile` instead?

src/xlFileGCN.c
```diff
@@ -0,0 +1,149 @@
...
+    nOffset = pFile->nOffset;
+    nSize = pFile->nSize;
+    if (nOffset + nSizeBytes > nSize) {
+        nSizeBytes = nSize - nOffset;
+    }
+    if (nSizeBytes == 0) {
+        *(s8*)pTarget = 0xFF;
+        return 0;
+    }
+    for (nSizeBytes; nSizeBytes != 0;) {
```
cadmic, line 92: `while (nSizeBytes != 0)`?

src/xlFileGCN.c
```diff
@@ -0,0 +1,149 @@
...
+        nSizeUsed = nSizeBytes;
+        if (nSizeUsed > 0x1000) {
+            nSizeUsed = 0x1000;
+        }
+        nOffset = pFile->nOffset & 0xFFFFFFFC;
+        nOffsetExtra = pFile->nOffset & 0x3;
+        nSize = (nSizeUsed + nOffsetExtra + 0x1F) & 0xFFFFFFE0;
+        if (gpfRead != (void*)0) {
```
cadmic, line 100: `NULL` (ditto below)

src/xlObject.c
```diff
@@ -0,0 +1,146 @@
...
+
+        if (xlListTestItem(gpListData, pData) && pData->pType == pType) {
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
+inline s32 xlObjectFindType(void* pObject, _XL_OBJECTTYPE* pType) {
```
cadmic, line 94: per your PR description, does this work as `_XL_OBJECTTYPE* ppType` and passing in `&pType`?

src/xlFileGCN.c
Yanis42: unfortunately no

src/xlObject.c
Yanis42: doesn't seem to work

cadmic (approved):

PR #53, opened 2024-03-02
Match cpuExecuteOpcode and some MIPS-related functions

cadmic:
`cpuExecuteOpcode` is the main MIPS interpreter if the MIPS code is not recompiled. The most interesting thing I found was a hack for `frameGetDepth` when interpreting `lhu`

### DIFF
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -3616,11 +3616,11 @@ static s32 cpuExecuteOpcode(Cpu* pCPU, s32 nCount0, s32 nAddressN64, s32 nAddres
             nAddress = pCPU->aGPR[MIPS_RS(nOpcode)].s32 + MIPS_IMM_S16(nOpcode);
             if (CPU_DEVICE(nAddress)->pfGet32(CPU_DEVICE(nAddress)->pObject,
                                               nAddress + CPU_DEVICE(nAddress)->nOffsetAddress, &nData32)) {
-                if (MIPS_FT(nOpcode) & 1) {
-                    pCPU->aFPR[MIPS_FT(nOpcode) - 1].u64 &= 0xFFFFFFFF;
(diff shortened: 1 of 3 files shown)


src/cpu.c
```diff
@@ -11,6 +13,30 @@
...
+//  float: opcode (6 bits) | fmt (5 bits) | ft (5 bits) | fs (5 bits) | fd (5 bits) | funct (6 bits)
+#define MIPS_OP(inst) ((inst) >> 26)
+#define MIPS_RS(inst) (((inst) >> 21) & 0x1F)
+#define MIPS_RT(inst) (((inst) >> 16) & 0x1F)
+#define MIPS_RD(inst) (((inst) >> 11) & 0x1F)
+#define MIPS_SA(inst) (((inst) >> 6) & 0x1F)
+#define MIPS_FUNCT(inst) ((inst)&0x3F)
+#define MIPS_IMM_S16(inst) ((s16)((inst)&0xFFFF))
+#define MIPS_IMM_U16(inst) ((u16)((inst)&0xFFFF))
+#define MIPS_TARGET(inst) ((inst)&0x3FFFFFF)
```
Yanis42, line 29:
```suggestion
#define MIPS_FUNCT(inst) ((inst) & 0x3F)
#define MIPS_IMM_S16(inst) ((s16)((inst) & 0xFFFF))
#define MIPS_IMM_U16(inst) ((u16)((inst) & 0xFFFF))
#define MIPS_TARGET(inst) ((inst) & 0x3FFFFFF)
```

Yanis42 (commented):
just a tiny detail

src/cpu.c
cadmic: This is actually clang-format's doing, and I can't find a way to turn it off

Yanis42: that's odd lol but ok

Yanis42 (approved):

PR #54, opened 2024-03-03
Match some CPU functions

Yanis42:
<details>
<summary>Functions</summary>

- cpuExecuteJump
- cpuFreeDevice
- cpuMapAddress
- cpuGetMode
- cpuGetSize
- __cpuERET
- __cpuBreak
- cpuSetXPC
- cpuEvent
- cpuGetAddressOffset
- cpuGetOffsetAddress
- cpuInvalidateCache
- cpuGetFunctionChecksum
- cpuHeapReset
- cpuHeapTake
- cpuDMAUpdateFunction
</details>

also thanks to cadmic for the help


src/cpu.c
```diff
@@ -577,6 +578,8 @@ const f64 D_80135FB0 = 3.0;
 const f32 D_80135FB8 = 0.5f;
 const f64 D_80135FC0 = 4503601774854144.0;
 
+static s32 cpuDMAUpdateFunction(Cpu* pCPU, s32 start, s32 end);
```
cadmic, line 581: I've been putting these function declarations a bit earlier in the file, and I don't really care where they go but it would be nice if they were all in the same place

src/cpu.c
```diff
@@ -4132,27 +4156,156 @@ inline s32 cpuMakeCachedAddress(Cpu* pCPU, s32 nAddressN64, s32 nAddressHost, Cp
 
 #pragma GLOBAL_ASM("asm/non_matchings/cpu/cpuException.s")
 
+#ifndef NON_MATCHING
+static s32 cpuMakeDevice(Cpu* pCPU, s32* piDevice, void* pObject, s32 nOffset, u32 nAddress0, u32 nAddress1, s32 nType);
 #pragma GLOBAL_ASM("asm/non_matchings/cpu/cpuMakeDevice.s")
+#endif
 
-#pragma GLOBAL_ASM("asm/non_matchings/cpu/cpuFreeDevice.s")
+s32 cpuFreeDevice(Cpu* pCPU, s32 i) {
```
cadmic, line 4164: rename i -> iDevice and j -> iAddress?

Also does early return work instead of `ret`?

src/cpu.c
```diff
@@ -4132,27 +4156,156 @@ inline s32 cpuMakeCachedAddress(Cpu* pCPU, s32 nAddressN64, s32 nAddressHost, Cp
...
+    }
+
+    return 1;
+}
 
 #pragma GLOBAL_ASM("asm/non_matchings/cpu/cpuSetTLB.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/cpu/cpuGetMode.s")
+static s32 cpuGetMode(u64 nStatus, CpuMode* peMode) {
+    if (nStatus & (1 << 1)) {
```
cadmic, line 4225: nit: I would just write `2` and `4` for these flags for now, I think it's more consistent with how we've been doing it (e.g. the `nStatus & 0x18` below)

src/cpu.c
```diff
@@ -4178,11 +4331,87 @@ s32 cpuSetCodeHack(Cpu* pCPU, s32 nAddress, s32 nOpcodeOld, s32 nOpcodeNew) {
...
+
+    for (i = 0; i < ARRAY_COUNT(pCPU->apDevice); i++) {
+        pCPU->apDevice[i] = NULL;
+    }
+
+    return 1;
+}
+
+inline s32 cpuFreeAllDevices(Cpu* pCPU) {
+    s32 i = 0;
```
cadmic, line 4358: nit: redundant `i = 0`

src/cpu.c
```diff
@@ -4197,23 +4426,348 @@ s32 cpuGetAddressBuffer(Cpu* pCPU, void** ppBuffer, u32 nAddress) {
...
+                pFunction->heapID = 3;
+                pFunction->heapWhere = -1;
+                if (!xlHeapTake(heap, memory_size)) {
+                    return 0;
+                }
+                return 1;
+            }
+        } else if (pFunction->heapID == 2) {
+            pFunction->heapID = 2;
+            nBlockCount = (memory_size + 0x9FF) / 2560;
```
cadmic, line 4584: write as 0xA00 (and 512 -> 0x200) for symmetry with the last couple lines?

src/cpu.c
```diff
@@ -4197,23 +4426,348 @@ s32 cpuGetAddressBuffer(Cpu* pCPU, void** ppBuffer, u32 nAddress) {
...
+        }
+
+        return 1;
+    }
+
+    if (pFunction->heapWhere == -1) {
+        return 0;
+    }
+
+    nMask = ((1 << (pFunction->heapWhere >> 0x10)) - 1) << (pFunction->heapWhere & 0x1F);
```
cadmic, line 4668: nit: I'd probably write `>> 16` although I don't feel strongly about it

src/cpu.c
```diff
@@ -4197,23 +4426,348 @@ s32 cpuGetAddressBuffer(Cpu* pCPU, void** ppBuffer, u32 nAddress) {
...
+    s32 iPack;
+    u32 nPack;
+    u32 nMask;
+    u32 nMask0;
+
+    done = 0;
+    for (iPack = 0; iPack < 125; iPack++) {
+        if ((nPack = aHeapTreeFlag[iPack]) != -1) {
+            nMask = 1;
+            nOffset = 1 << 5;
```
cadmic, line 4694: 32 instead of `1 << 5` here and below?

src/cpu.c
Yanis42: yea it works

src/cpu.c
Yanis42: I think ``>> 16`` is more consistent

cadmic (approved):

PR #57, opened 2024-03-04
Match cpuExecuteLoadStore and cpuExecuteLoadStoreF

cadmic:
These recompile load/store instructions into PowerPC machine code. I've given up on annotating the PPC instructions for now, ideally we'd make macros or inline functions for them but it's not fully clear what those look like yet (e.g. when to use `|` vs `+` for immediates)

### DIFF
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -3832,6 +3832,7 @@ static s32 cpuExecuteCall(Cpu* pCPU, s32 nCount, s32 nAddressN64, s32 nAddressGC
     return nAddressGCN;
 }
 
+// Matches but data doesn't
 #ifndef NON_MATCHING
 static s32 cpuExecuteLoadStore(Cpu* pCPU, s32 nCount, s32 nAddressN64, s32 nAddressGCN);
 #pragma GLOBAL_ASM("asm/non_matchings/cpu/cpuExecuteLoadStore.s")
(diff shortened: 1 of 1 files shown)


src/cpu.c
```diff
@@ -3826,11 +3832,499 @@ static s32 cpuExecuteCall(Cpu* pCPU, s32 nCount, s32 nAddressN64, s32 nAddressGC
...
+    u8 device;
+    s32 total;
+    s32 count;
+    s32 save;
+    s32 interpret;
+    s32* before;
+    s32* after;
+    s32 check2;
+    s32* anCode;
+    s32 pad;
```
Yanis42, line 3853: the stack macro doesn't work here?

src/cpu.c
```diff
@@ -3826,11 +3832,499 @@ static s32 cpuExecuteCall(Cpu* pCPU, s32 nCount, s32 nAddressN64, s32 nAddressGC
...
+    s32 total;
+    s32 count;
+    s32 save;
+    s32 interpret;
+    s32* before;
+    s32* after;
+    s32 check2;
+    s32* anCode;
+    s32 rt;
+    s32 pad;
```
Yanis42, line 4130: same thing here

Yanis42 (commented):
are these functions not matching because of data stuff? if it's the reason you forgot to put the comment about it like we're doing usually

src/cpu.c
cadmic: it does, but I haven't been using unless necessary (there's lots of other `s32 pad;` in this file already). Should we prefer using the macro here? One downside of the PAD_STACK() macro is that it doesn't work if you need to insert some padding in between local variables so we can't use it in all cases.

cadmic:
Yeah, they have jump tables (I added the comment)

src/cpu.c
Yanis42: I think it's weird to have a macro for this purpose and not using it but that can be cleaned up later

Yanis42 (approved):

PR #62, opened 2024-03-25
Match library "infrastructure"

cadmic:
Basically the parts that aren't implementations of library functions


Yanis42 (approved):
it's interesting that ``libraryUpdate`` became inlined on CE

PR #63, opened 2024-03-25
Match guPerspective[F]

cadmic:
Relevant to gz. Note these functions call `frameSetMatrixHint` to alert the graphics system and that the `scale` argument is ignored unlike the N64 implementation

### DIFF
--- a/src/library.c
+++ b/src/library.c
@@ -1,5 +1,8 @@
 #include "library.h"
+#include "frame.h"
 #include "library_jumptables.h"
+#include "math.h"
+#include "system.h"
 
 char D_800EEB00[] = "OS-LIBRARY";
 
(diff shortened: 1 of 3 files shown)


src/library.c
```diff
@@ -1343,9 +1346,166 @@ const f32 D_801360D4 = -2.0;
...
+        for (j = 0; j < 4; j++) {
+            if (i == j) {
+                mf[i * 4 + j] = data1.u32;
+            } else {
+                mf[i * 4 + j] = data0.u32;
+            }
+        }
+    }
+
+    fovy *= 3.1415926f / 180;
```
Yanis42, line 1403: I think this should be a macro ``#define PI 3.1415926f``

Yanis42 (commented):
tiny thing otherwise it's fine

src/library.c
cadmic: changed to use the stdlib `M_PI`

Yanis42 (approved):

PR #64, opened 2024-03-29
Match Dolphin SDK - OS files

Yanis42:
this also adds new headers and update others and I had to make few changes in the Makefile since the Dolphin SDK is using a different compiler version and different flags, also I decided to move the emulator's files in an "emulator" folder, and the sdk files to a "dolphin" folder, and sorry big PR

---
These files were matched thanks to the following projects:
- Dolphin SDK decompilation: https://github.com/doldecomp/dolsdk2001 (and [camthehaxman's ``r37`` branch](https://github.com/camthehaxman/dolsdk2001/tree/r37))
- Metroid Prime: https://github.com/PrimeDecomp/prime
- Mario Kart Double Dash: https://github.com/SwareJonge/mkdd
- Pikmin 2: https://github.com/projectPiki/pikmin2


Makefile
```diff
@@ -61,16 +67,20 @@ ASM_PROCESSOR := $(ASM_PROCESSOR_DIR)/compile.sh
 POSTPROC := tools/postprocess.py
 
 # Options
-INCLUDES := -Iinclude
+INCLUDES := -Iinclude -i libc
```
cadmic, line 70: Why `-I` for include but `-i` for libc?

Makefile
```diff
@@ -61,16 +67,20 @@ ASM_PROCESSOR := $(ASM_PROCESSOR_DIR)/compile.sh
...
+ASFLAGS := -mgekko -I include -I libc
 
 # Linker Flags
-LDFLAGS := -map $(MAP) -fp hard -nodefaults -w off
+LDFLAGS := -map $(MAP) -fp hardware -nodefaults -warn off
 
 # Compiler Flags
-CFLAGS := -Cpp_exceptions off -proc gekko -fp hard -fp_contract on -enum int -O4,p -inline auto,deferred -sym on -nodefaults -msgstyle gcc $(INCLUDES)
+CFLAGS := -Cpp_exceptions off -proc gekko -fp hard -fp_contract on -enum int -O4,p -sym on -nodefaults -msgstyle gcc $(INCLUDES) -DDOLPHIN_REV=$(DOLPHIN_REVISION)
+DOLPHIN_SDK_CFLAGS := $(CFLAGS) -align powerpc -maxerrors 1 -nosyspath -RTTI off -str reuse -multibyte -inline auto
```
cadmic, line 80: should any of these flags be set on the emulator too?

Makefile
```diff
@@ -51,6 +54,9 @@ MWCC_DIR := tools/mwcc_compiler/$(MWCC_VERSION)
 CC := $(WINE) $(MWCC_DIR)/mwcceppc.exe
 LD := $(WINE) $(MWCC_DIR)/mwldeppc.exe
 
+DOLPHIN_SDK_MWCC_DIR := tools/mwcc_compiler/GC/1.2.5n
+DOLPHIN_SDK_CC := $(WINE) $(DOLPHIN_SDK_MWCC_DIR)/mwcceppc.exe
```
cadmic, line 58: Should this be `DOLPHIN_` instead of `DOLPHIN_SDK_`? e.g. the directories are called `dolphin`

Makefile
```diff
@@ -136,8 +146,14 @@ $(DOL): $(ELF)
...
+
+$(BUILD_DIR)/src/emulator/%.o: src/emulator/%.c
 	$(ASM_PROCESSOR) "$(CC) $(CFLAGS)" "$(AS) $(ASFLAGS)" $@ $<
 
-$(BUILD_DIR)/%.o: %.cpp
+$(BUILD_DIR)/src/dolphin/%.o: src/dolphin/%.cpp
+	$(DOLPHIN_SDK_CC) $(DOLPHIN_SDK_CFLAGS) -c -o $@ $<
+
+$(BUILD_DIR)/src/emulator/%.o: src/emulator/%.cpp
 	$(CC) $(CFLAGS) -c -o $@ $<
```
cadmic, line 159: should we just delete these rules?

libc/stdlib.h
cadmic, line 1: should `atoi` be here? Then we could get rid of the TODO in system.c

include/library.h
cadmic, line 1: should we organize the headers in the same way, e.g. `include/emulator` too?

include/xlHeap.h
```diff
@@ -13,5 +13,6 @@ s32 xlHeapFill32(void* pHeap, s32 nByteCount, u32 nData);
 s32 xlHeapGetFree(s32* pnFreeBytes);
 s32 xlHeapSetup(void* pHeap, s32 nSizeBytes);
 s32 xlHeapReset(void);
+s32 xlHeapBlockCacheAdd(u32* pBlock);
```
cadmic, line 16: nit: ideally this should be in the same order as the C file (although why does this need to be declared at all?)

include/library.h
```diff
@@ -82,6 +82,7 @@ void zeldaLoadSZS_Entry(Cpu* pCPU);
 void zeldaLoadSZS_Exit(Cpu* pCPU);
 
 s32 libraryTestFunction(Library* pLibrary, CpuFunction* pFunction);
+s32 libraryUpdate(Library* pLibrary);
```
cadmic, line 85: I think it would be easier if we don't declare this in the header, nothing outside of library.c needs it and it's inlined away in CE

include/dolphin/os/OSAlarm.h
```diff
@@ -18,8 +22,18 @@ struct OSAlarm {
 };
 
 void OSInitAlarm(void);
-void OSCreateAlarm(OSAlarm* alarm);
 void OSSetAlarm(OSAlarm* alarm, OSTime tick, OSAlarmHandler handler);
+void OSSetAlarmTag(OSAlarm* alarm, u32 tag);
```
cadmic, line 26: Many of these functions aren't defined in OOT. Does it still make sense to declare them in the headers?

I could see two philosophies here:
* Only declare functions that are actually defined (i.e. how the headers were before this change). That way, when you're modding you get a compile error and not a link error if you try to use a function that has been removed from OoT
* Declare all functions in the SDK, including removed ones. That way headers are more consistent with some upstream (but what is the upstream here?)

src/dolphin/base/PPCArch.c
```diff
@@ -0,0 +1,195 @@
...
+        u32 fpscr_pad;
+        u32 fpscr;
+    } u;
+};
+
+#define HID0_SPD 0x00000200 // Speculative cache access enable (0 enable)
+
+void PPCMthid0(u32 newHID0);
+
+ASM u32 PPCMfmsr(void){
```
cadmic, line 17: format?

Makefile
Yanis42: I wanted to stay consistent with what dtk is using to make sure it works, I didn't think of making it consistent afterwards (I changed that)

Makefile
Yanis42: I didn't want to add more flags for the emulator's files because I have no idea what they are doing/how code can be affected

Yanis42: also sdk flags are missing 2 ``-pragma`` because the asmprocessor is angry with ``"`` apparently

Makefile
```diff
@@ -35,6 +35,9 @@ include obj_files.mk
 
 MWCC_VERSION := GC/1.1
 
+# 58 for MQ, 99 for unknown
+DOLPHIN_REVISION := 99
```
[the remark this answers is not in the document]
Yanis42: in code I'm using 58 since it's the only revision we know, I just wanted to use a higher number than 58 as CE is using an updated revision

maybe we could use the build year instead? so 2002 and 2003? I'm not sure if this is a good idea for the other versions we'll support eventually though

Makefile
Yanis42: yeah I was thinking the same thing earlier, afaik there's no C++ in this codebase

include/xlHeap.h
Yanis42: I think this is just something I had to add for MQ, I'll remove it since this version isn't supported yet anyway

include/library.h
Yanis42: yeah good idea, we could also get a global ``emulator.h`` but idk if it's worth

Makefile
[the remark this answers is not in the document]
Yanis42: I decided to use the build year, I think it should be fine for the other versions anyway as they should share the same revision afaik

include/dolphin/os/OSAlarm.h
Yanis42: it took some time but I removed the declarations for the functions that aren't defined

Makefile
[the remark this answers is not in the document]
cadmic: What define would we use to compile MQ vs CE in the emulator code? What do you think about using the same one here?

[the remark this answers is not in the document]
Yanis42: for the emu's code, for US and JP regions it will probably be the same code (more or less), but PAL versions are a bit different so I don't think the build date is enough, maybe we can use an enum and use ``-DVERSION=1`` for MQ-US for instance and have ``#if VERSION == MQ_US`` with MQ_US being an enum member

but that's just a small idea I just had, not sure if that's ideal

src/dolphin/os/OS.c
```diff
@@ -0,0 +1,655 @@
...
+        return OS_CONSOLE_ARTHUR;
+    }
+    return BootInfo->consoleType;
+}
+
+void* __OSSavedRegionStart;
+void* __OSSavedRegionEnd;
+
+extern u32 BOOT_REGION_START AT_ADDRESS(0x812FDFF0); //(*(u32 *)0x812fdff0)
+extern u32 BOOT_REGION_END AT_ADDRESS(0x812FDFEC); //(*(u32 *)0x812fdfec)
```
cadmic, line 161: nit: remove comments?

src/dolphin/os/OS.c
```diff
@@ -0,0 +1,655 @@
...
+
+void ClearArena(void) {
+    if ((u32)(OSGetResetCode() + 0x80000000) != 0U) {
```
cadmic, line 164: `OSGetResetCode() != 0x80000000`?

src/dolphin/os/OS.c
```diff
@@ -0,0 +1,655 @@
...
+        if (BootInfo == NULL || (inputConsoleType = BootInfo->consoleType) == 0) {
+            inputConsoleType = OS_CONSOLE_ARTHUR; // default console type
+        } else {
+            inputConsoleType = BootInfo->consoleType;
+        }
+
+// work out what console type this corresponds to and report it
+// consoleTypeSwitchHi = inputConsoleType & 0xF0000000;
+#if DOLPHIN_REV == 2002
+        switch (inputConsoleType & 0xffff0000)
```
cadmic, line 322: this 0x0000ffff vs 0x0fffffff (or its inverse 0xffff0000 vs 0xf0000000) appears multiple times here, maybe it would be cleaner to give it a name

cadmic (commented):
I only skimmed the .c files and didn't look in detail, but if it matches it matches

include/dolphin/os/OSAlarm.h
Thar0: Personally I prefer the second option, where upstream is ideally a dedicated SDK decompilation with complete headers.

I think a link error is actually better than a compile error caused by incomplete headers, since at that point the fix is just to source the function's implementation rather than source both the implementation and a sufficiently complete header to use with it.

src/dolphin/os/OS.c
Yanis42: I called them "HI_MASK" and "LO_MASK", I'm not sure if these names are good or not though

include/dolphin/os/OSAlarm.h
Yanis42: that's something we can do later as cleanup work imo, I just wanted to match these files and I wasn't really motivated to add every missing declarations

but yeah ideally in the future this will be done

Makefile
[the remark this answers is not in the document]
cadmic: I think something like `-DVERSION=MQ_US` would work too, but this could be easily changed later so I think it's fine as-is for now

cadmic (approved):

PR #67, opened 2024-04-13
Match matrix library functions

cadmic:
Mostly straightforward but had to add some `volatile` hacks in `guMtxIdentF` and `guMtxF2L`


Yanis42 (approved):
I don't have much to say about it as it's math stuff but it builds properly so it's fine for me

PR #68, opened 2024-04-14
Match libultra OS threading library functions

cadmic:
plus fixup `CPU_DEVICE` macros and function signatures. `__osException` seems to require some `goto`s, there might be a way around it but it's not too surprising since the N64 function it's trying to reimplement is handwritten assembly


src/emulator/library.c
```diff
@@ -330,27 +341,779 @@ const f64 D_801360C8 = 4503601774854144.0;
...
+    /* 0x4 */ OSMesgQueue_s* retQueue;
+} OSIoMesgHdr_s; // size = 0x8
+
+typedef struct OSIoMesg_s {
+    /* 0x00 */ OSIoMesgHdr_s hdr;
+    /* 0x08 */ void* dramAddr;
+    /* 0x0C */ u32 devAddr;
+    /* 0x10 */ u32 size;
+    /* 0x14 */ void* piHandle;
+} OSIoMesg_s; // size = 0x18
```
Yanis42, line 445: maybe you should move these structs to their own header, like ``library_os.h`` or something

src/emulator/library.c
```diff
@@ -330,27 +341,779 @@ const f64 D_801360C8 = 4503601774854144.0;
...
+    if (!cpuGetRegisterCP0(pCPU, 13, &nCause)) {
+        return 0;
+    }
+
+    nData32 = nCause;
+    __osRunningThread->context.cause = nData32;
+
+    __osRunningThread->state = 2;
+    switch ((nCause & 0x7C) >> 2) {
+        case 0: // interrupt
```
Yanis42, line 574: is it worth to make enums for these?

Yanis42 (commented):
I'm not a fan of gotos but I guess we don't have the choice in this case

src/emulator/library.c
cadmic: Turns out there already is one :)

src/emulator/library.c
cadmic: Hmm there's also some structs for the gu/`Hilite` stuff below so if we move these then we should probably move those too. TBH I don't mind them being in the C file, they're "implementation details" that other files will not need to know about and they're not horribly big

Yanis42 (approved):
nice work! also congrats for matching the other functions of library.c

PR #69, opened 2024-04-15
Finish library.c

cadmic:
Matches the remaining functions, and the data. `gaFunction` had to be moved for data ordering, which means most of the forward declarations aren't necessary so I moved them out of the header since these functions not meant to be public.


Yanis42 (approved):
feels nice to see the asm folder having files removed 👀

PR #72, opened 2024-04-18
Match frame matrix functions and some framebuffer stuff

cadmic:
I gave up on `frameSetMatrix` and just slapped some `volatile`s on the problem

### DIFF
--- a/src/emulator/frame.c
+++ b/src/emulator/frame.c
@@ -1,5 +1,5 @@
-#include "dolphin.h"
 #include "emulator/frame.h"
+#include "dolphin.h"
 #include "emulator/ram.h"
 #include "emulator/rsp.h"
 #include "emulator/simGCN.h"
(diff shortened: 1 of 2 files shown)


src/emulator/frame.c
```diff
@@ -665,7 +877,85 @@ s32 frameBeginOK(void) {
 
 #pragma GLOBAL_ASM("asm/non_matchings/frame/frameHackCIMG_Zelda2.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/frame/frameHackCIMG_Zelda.s")
+s32 frameHackCIMG_Zelda(Frame* pFrame, FrameBuffer* pBuffer, u64* pnGBI, u32 nCommandLo) {
```
Yanis42, line 880: just out of curiosity, what is this doing exactly?

src/emulator/frame.c
```diff
@@ -685,6 +975,26 @@ s32 frameBeginOK(void) {
 
 #pragma GLOBAL_ASM("asm/non_matchings/frame/frameEvent.s")
 
+inline s32 frameCopyMatrix(Mtx44 matrixTarget, Mtx44 matrixSource) {
```
Yanis42, line 978: loops doesn't match here?

Yanis42 (approved):
nice!

src/emulator/frame.c
cadmic: Doesn't seem like it but maybe there's a way

src/emulator/frame.c
cadmic: I believe it copies the framebuffer into N64 memory whenever the game needs it (pausing or the equipment screen). It's why I was interested, because of defeat ganon

Yanis42: ooooooooooooohhhh that's THE function then 👀 👀 👀 👀 👀

PR #77, opened 2024-04-25
Run host compiler for better warnings/errors (CC_CHECK)

cadmic:
Inspired by the OOT repo, mainly to enforce that functions are declared before they are used because that's been a common footgun. It also found that stuff was duplicated between `dolphin/exi.h` and `dolphin/os/OSExpansion.h` so I merged the two but I'm not sure if this is correct.


cadmic:
Tested with GCC, although there's still some unrecognized warning flags (although that's merely annoying, not critical)

cadmic:
I rebased, removed clang-specific flags, and turned off `-Werror`. GCC gives a warning for "used but not defined" if a static function is not yet decompiled which is annoying for partially-complete files, and there's no way to turn it off. But maybe it's still worth it?

Yanis42 (approved):

PR #81, opened 2024-04-28
Match most of the THP library

Yanis42:
not including the SDK files, THPRead has one non-matching, the other things are data issues I couldn't figure out


Makefile
```diff
@@ -154,5 +154,11 @@ $(BUILD_DIR)/%.o: %.s
 $(BUILD_DIR)/src/dolphin/%.o: src/dolphin/%.c
 	$(ASM_PROCESSOR) "$(DOLPHIN_CC) $(DOLPHIN_CFLAGS)" "$(AS) $(ASFLAGS)" $@ $<
 
+$(BUILD_DIR)/src/emulator/THPRead.o: src/emulator/THPRead.c
+	$(ASM_PROCESSOR) "$(CC) $(EMULATOR_CFLAGS)" "$(AS) $(ASFLAGS)" $@ $<
+
+$(BUILD_DIR)/src/emulator/THP%.o: src/emulator/THP%.c
+	$(ASM_PROCESSOR) "$(CC) $(DOLPHIN_CFLAGS)" "$(AS) $(ASFLAGS)" $@ $<
```
cadmic, line 161: This is fun. I don't think the name `DOLPHIN_CFLAGS` makes much sense here. What do you think about getting rid of it and setting `-inline deferred` explicitly?
```
CFLAGS := -Cpp_exceptions off -proc gekko -fp hardware -fp_contract on -enum int -O4,p -inline auto -sym on -nodefaults -msgstyle gcc $(INCLUDES) -DDOLPHIN_REV=$(DOLPHIN_REVISION)

...


$(BUILD_DIR)/src/dolphin/%.o: src/dolphin/%.c
	$(ASM_PROCESSOR) "$(DOLPHIN_CC) $(CFLAGS)" "$(AS) $(ASFLAGS)" $@ $<

$(BUILD_DIR)/src/emulator/THPRead.o: src/emulator/THPRead.c
	$(ASM_PROCESSOR) "$(CC) $(CFLAGS) -inline deferred" "$(AS) $(ASFLAGS)" $@ $<

$(BUILD_DIR)/src/emulator/THP%.o: src/emulator/THP%.c
	$(ASM_PROCESSOR) "$(CC) $(CFLAGS)" "$(AS) $(ASFLAGS)" $@ $<

$(BUILD_DIR)/src/emulator/%.o: src/emulator/%.c
	$(ASM_PROCESSOR) "$(CC) $(CFLAGS) -inline deferred" "$(AS) $(ASFLAGS)" $@ $<
```

src/emulator/THPRead.c
```diff
@@ -52,30 +60,576 @@ const f32 D_80135DCC = 160.0;
...
+
+        gMovieErrorToggle = 0;
+        readBuffer->frameNumber = readFrame;
+        PushReadedBuffer(readBuffer);
+        offset += size;
+        size = *(s32*)readBuffer->ptr;
+
+        frameNumber = (readFrame + ActivePlayer.initReadFrame) % ActivePlayer.header.numFrames;
+        if (frameNumber == ActivePlayer.header.numFrames - 1) {
+            if ((ActivePlayer.playFlag & 1))
```
cadmic, line 599: maybe we should set `InsertBraces: true` in the clang-format config

src/emulator/THPPlayer.c
```diff
@@ -43,37 +30,847 @@ static s16* LastAudioBuffer;
...
+        DCFlushRange(SoundBuffer, sizeof(SoundBuffer));
+        AIInitDMA((u32)&SoundBuffer[SoundBufferIndex], 320 * 2);
+        AIStartDMA();
+    }
+
+    Initialized = TRUE;
+    return TRUE;
+}
+
+#ifdef UNUSED
```
cadmic, line 82: keeping unused functions is interesting, but maybe it's a little weird since we don't do that with the emulator or dolphin code. Like do we need to commit this code if it's not part of reverse-engineering the emulator?

Makefile
Yanis42: I wasn't sure what to do for that, it's odd that only THPRead was compiled with deferred

src/emulator/THPPlayer.c
Yanis42: I wasn't sure what would be best, I'll just remove them 👍

src/emulator/THPRead.c
Yanis42: I tried but it seems clang-format doesn't recognize this rule for me

Makefile
cadmic: I mean, I think we should get rid of `DOLPHIN_CFLAGS` and `EMULATOR_CFLAGS` entirely (and move `-inline auto` to `CFLAGS`) since the only difference seems to be `-inline deferred`, and now it seems that's not a dolphin/emulator thing because of the THP files

src/emulator/THPRead.c
cadmic: Alright, we can leave it alone for now

Makefile
Yanis42: I did the change, let me know if that's what you had in mind

cadmic: thanks, it's close enough

PR #82, opened 2024-04-29
Mostly match THPRead data

cadmic:
Still some issues with `static` which maybe will become clearer once simGCN code is matched. But most of it is pure C now at least


include/emulator/simGCN.h
```diff
@@ -80,6 +80,21 @@ extern System* gpSystem;
...
+extern u8 gcoverOpen[];
+extern u8 gnoDisk[];
+extern u8 gretryErr[];
+extern u8 gfatalErr[];
+extern u8 gwrongDisk[];
+extern u8 greadingDisk[];
+extern u8 gbar[];
+extern u8 gyes[];
+extern u8 yno[];
+extern u8 gmesgOK[];
```
Yanis42, line 92: is there really no way we can use ``TEXPalette`` on these ones? no big deal if we can't, I was just wondering

Yanis42 (approved):

include/emulator/simGCN.h
cadmic: There might be a way, I'm not sure. It's kinda weird though, the bytearrays don't have real pointers in them (just offsets in to the byte array for things like `descriptorArray`) but then `simulatorUnpackTexPalette` fixes them all in-place (https://decomp.me/scratch/iugKY). So they don't really start off as `TEXPalette`s but become them later I guess, and I'm not sure how to initialize that in C

PR #87, opened 2024-05-01
boolify THP files

cadmic:
also fix some movieReset inlines

### DIFF
--- a/src/emulator/THPRead.c
+++ b/src/emulator/THPRead.c
@@ -20,15 +20,15 @@ static f32 gOrthoMtx[4][4];
 //! TODO: make static (data ordering issues)
 // and remove suffix (there's other global variables sharing the same name)
 u32 gnTickReset_thpread;
-s32 gbReset_thpread;
-s32 toggle_184;
+bool gbReset_thpread;
+bool toggle_184;
 
-s32 gMovieErrorToggle;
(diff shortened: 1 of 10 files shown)


src/emulator/THPRead.c
```diff
@@ -329,73 +327,48 @@ s32 movieDVDShowError(s32 nStatus, void*, s32, u32) {
...
-            gnTickReset_thpread = nTick;
-            if ((gbReset_thpread == 0) || (bFlag != 0U)) {
-                gbReset_thpread = bFlag;
-            } else {
-                movieReset(1, 0);
-            }
-        } else if ((nTick - gnTickReset_thpread) >= OSSecondsToTicks(0.5f)) {
-            movieReset(1, 0);
-        }
+        movieTestReset(true, false);
```
Yanis42, line 340: this is suspicious, isn't this supposed to be a switch? it would make more sense for that whole part but idk if it would match

Yanis42 (approved):

src/emulator/THPRead.c
Yanis42: (didn't realised it would take the removed lines)

cadmic: yeah it is weird, I tried a few things but it doesn't seem to match

PR #89, opened 2024-05-02
Match paired-single/quantization (`psq`) functions

cadmic:
Background on the GameCube CPU's unique `psq` instructions: https://mariokartwii.com/showthread.php?tid=1870

GQRs 2-5 are from OSFastCast.h. I tried a couple versions but ultimately the ones from the [open_rvl](https://github.com/kiwi515/open_rvl/blob/main/include/revolution/OS/OSFastCast.h) repo worked the best.

GQRs 6-7 appear to be custom, for decoding 1.7 and 11.5 fixed-point numbers used by the GBI. They're initialized in `xlMain()` and only used in frame.c. I'm not sure what to call them or where to put them so I left a TODO to think about it later.

Finally, `PSMTX44MultVecNoW` looks to be handwritten and it's probably the cause of `peephole off` in frame.c

### DIFF
--- a/src/emulator/frame.c
+++ b/src/emulator/frame.c
@@ -2138,9 +2138,7 @@ bool frameGetMatrix(Frame* pFrame, Mtx44 matrix, FrameMatrixType eType, bool bPu
 
 // TODO: move these paired-single/quantization functions to a separate header
 // along with the GQR initialization in xlMain()?
-inline void s16tof32(register s16* in, register f32* out) {
-    OSs16tof32(in, out);
-}
+inline void s16tof32(register s16* in, register f32* out) { OSs16tof32(in, out); }
 
 inline void s16tof32Pair(register s16* in, register f32* out) {
 #ifdef __MWERKS__
@@ -2595,7 +2593,7 @@ bool frameSetMatrixHint(Frame* pFrame, FrameMatrixProjection eProjection, s32 nA
     }
 
     if (eProjection == 1) {
-        rScale = 0.0f;
+        rNear = 0.0f;
     }
 
     pFrame->aMatrixHint[iHint].nCount = 4;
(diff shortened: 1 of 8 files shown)


src/emulator/frame.c
```diff
@@ -2137,7 +2136,316 @@ bool frameGetMatrix(Frame* pFrame, Mtx44 matrix, FrameMatrixType eType, bool bPu
...
+    f32 rT;
+    f32 arNormal[3];
+    f32 arPosition[3];
+    Vertex* pVertex;
+    u32 nData32;
+    Light* aLight;
+    Light* pLight;
+    s32 iVertex1;
+    // f32 rScale;
+    // f32 rScaleST;
```
Yanis42, line 2219: are these useful?

src/emulator/frame.c
```diff
@@ -2137,7 +2136,316 @@ bool frameGetMatrix(Frame* pFrame, Mtx44 matrix, FrameMatrixType eType, bool bPu
...
+                                      arPosition[2] * matrixView[2][2] + matrixView[3][2]);
+
+        if (nLight != 0) {
+            s8tof32Scaled128Pair(&pnData8[12], &arNormal[0]);
+            s8tof32Scaled128(&pnData8[14], &arNormal[2]);
+
+            iLight = nLight;
+            pLight = &aLight[iLight];
+            if (gpSystem->eTypeROM == SRT_STARFOX) {
+                while ((rColorR = pLight->rColorR) + (rColorG = pLight->rColorG) + (rColorB = pLight->rColorB) ==
```
Yanis42, line 2342: would a ``for`` match too? idk if it could help having something less weird like this

Yanis42 (commented):
nice, finally getting rid of the peephole pragma

src/emulator/frame.c
cadmic: Maybe, but based on the asm it seems like rColorR/rColorG/rColorB are assigned in the "test" part of the loop, so I don't think it would be any prettier

Yanis42 (approved):

PR #97, opened 2024-05-06
Match RSP update functions and rdpParseGBI

cadmic:
With this rdp.c is like 90% done

For the rdp commands, I used comments instead of enums for now because in the very similar rspParseGBI_F3DEX2, one switch statement is used to parse 3 different microcodes and I'm not sure how to make that work with enums yet.


Yanis42 (approved):
nice work!

PR #101, opened 2024-05-08
Match most functions of ``pif.c``

Yanis42:
3 functions left:
- pifContDataCrc
- pifProcessInputData
- pifProcessOutputData


src/emulator/pif.c
```diff
@@ -7,44 +10,361 @@ _XL_OBJECTTYPE gClassPIF = {
...
+        return false;
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/pif/pifPut8.s")
+    return true;
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/pif/pifPut16.s")
+bool pifPut8(Pif* pPIF, u32 nAddress, s8* pData) {
+    if ((nAddress & PIF_RAM_END) >= PIF_RAM_START) {
```
cadmic, line 152: I think these usages of `PIF_RAM_END` (or rather the name `PIF_RAM_END` for the constant 0x7FF here) is incorrect actually. If PIF RAM actually ended at 0x7EF or something we wouldn't want to use that to mask the address here. Something like `PIF_ADDRESS_MASK` maybe? Though I wouldn't mind seeing `0x7FF` inline either to make it immediately clear this is masking off everything but the low 11 bits

PR #103, opened 2024-05-09
Remove fake tXL_NODE struct

cadmic:
This doesn't exist according to the debug data, it was copied from oot-vc. And it's hard for m2c to deal with.


Yanis42 (approved):
nice

PR #106, opened 2024-05-09
Make decompme.py context parseable by m2c

cadmic:
The big issue here is that m2c can't handle most macros, but we don't want to run the C preprocessor because that would replace macros we want to keep too and strip all comments. So I hardcoded a list of macros to replace and manually replaced the ones in vitypes.h. (Although, maybe it's fine to just run the C preprocessor instead? Not having any macros in the context is pretty annoying though.)

m2c also can't handle inline asm, so those are now stripped too (by pretending that `__MWERKS__` is not defined) even though that could be important for matching.


include/dolphin/vi/vitypes.h
```diff
@@ -19,18 +19,18 @@
...
+    VI_TVMODE_NTSC_PROG = 2,
+    VI_TVMODE_PAL_INT = 4,
+    VI_TVMODE_PAL_DS = 5,
+    VI_TVMODE_EURGB60_INT = 20,
+    VI_TVMODE_EURGB60_DS = 21,
+    VI_TVMODE_MPAL_INT = 8,
+    VI_TVMODE_MPAL_DS = 9,
+    VI_TVMODE_DEBUG_INT = 12,
+    VI_TVMODE_DEBUG_PAL_INT = 16,
+    VI_TVMODE_DEBUG_PAL_DS = 17,
```
Yanis42, line 33: I think it would be nice to keep the ``VI_TVMODE`` thing as a comment

include/emulator/rsp.h
```diff
@@ -116,6 +117,19 @@ typedef struct __anon_0x58107 {
     /* 0x0 */ s16 anSlice[8];
 } __anon_0x58107; // size = 0x10
 
+typedef struct __anon_0x58360 {
+    /* 0x0 */ s16 r;
+    /* 0x2 */ s16 g;
+    /* 0x4 */ s16 b;
+    /* 0x6 */ s16 a;
+} __anon_0x58360; // size = 0x8
```
Yanis42, line 125: is this used in simGCN? the ``Colors_u32`` array

cadmic: I don't think so based on the debug data and the name

Yanis42 (approved):

PR #111, opened 2024-05-12
Write basic development guide

cadmic:
Still pretty bare-bones and probably only helpful for people already familiar with other decomp projects, but it's a start

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,15 +18,14 @@ You will need the following dependencies:
 * make
 * git
 * python3
-* WiBo (Linux)
-* 32-bit wine (macOS)
+* wibo (Linux) or wine (macOS)
 * wget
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,3 +50,41 @@ brew install coreutils python3 wget wine
...
+```
+mkdir expected
+cp -r build expected/
+```
+
+### Diff tools
+
+For locally diffing the current build against the expected build:
+
+* `objdump`: Start the objdump GUI and open this directory as the project directory.
```
Prakxo, line 67: Did you mean objdiff instead of objdump?

cadmic: oof, thank you

Yanis42 (approved):

PR #112, opened 2024-05-12
Match most functions of ``_buildtev.c``

Yanis42:
``SetupStage`` is the only function left (scratch: https://decomp.me/scratch/5ZLkZ)


src/emulator/_buildtev.c
```diff
@@ -30,24 +33,20 @@ static GXTevAlphaArg sUsualAArgs[] = {
     GX_CA_ZERO,
 };
 
-static s32 texelType[] = {
-    0x00000001, 0x00000002, 0x00000008, 0x00000009, 0x00000001, 0x00000002, 0x00000004, 0x00000002,
+static s32 texelType[][4] = {
```
cadmic, line 36: I think `texelType[2][4]` looks a little better

src/emulator/_buildtev.c
```diff
@@ -63,14 +62,225 @@ const s32 D_801360D8[2] = {
     0x00000000,
 };
 
-#pragma GLOBAL_ASM("asm/non_matchings/_buildtev/SetColor.s")
+//! TODO: move this once the MSL is matched
```
cadmic, line 65: I'm actually not sure about this, `memset` is called by name in other places. This seems to be a special memset only used in this file so far??

Yanis42: I updated the comment, let me know what you think about it

PR #116, opened 2024-05-13
Avoid using the asm processor on files that doesn't need it

Yanis42:
doing this saves ~3 seconds with ``make clean && make -j1``, also I'm not sure if there's a cleaner way to do this


Yanis42:
how do you check if a file contains something that requires the asmprocessor in bash?

cadmic:
maybe something like
```
if grep -q '^#pragma \(GLOBAL_ASM\|INCBIN\)'  "$C_FILE" then
    ...
fi
```

cadmic:
fixed by #127

PR #119, opened 2024-05-13
Set up permuter import

cadmic:
This assumes `wibo` because with `wine` it's kind of uselessly slow and crashy, at least on mac.

Testing:
```
# edit THPRead.c to mark movieDVDRead as inline
decomp-permuter/import.py src/emulator/THPRead.c asm/non_matchings/THPRead/Reader.s
decomp-permuter/permuter.py nonmatchings/Reader -j 8
```

Note you need a recent decomp-permuter to pick up https://github.com/simonlindholm/decomp-permuter/pull/170

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ __pycache__/
 
 # IDE artifacts
 .vscode/
+
+# Permuter artifacts
+nonmatchings/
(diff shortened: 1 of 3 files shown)


Yanis42 (approved):
looks good to me, imo you should add something to explain how to execute ``decomp-permuter/permuter.py``, this PR's description was very helpful in that regard

cadmic:
how does it look now?

Yanis42 (approved):

PR #121, opened 2024-05-14
Match some more frame functions

cadmic:
Unfortunately tons of `#ifndef NON_MATCHING` because of floats

### DIFF
--- a/src/emulator/frame.c
+++ b/src/emulator/frame.c
@@ -2701,8 +2701,25 @@ static inline bool frameCopyMatrix(Mtx44 matrixTarget, Mtx44 matrixSource) {
     return true;
 }
 
-bool frameScaleMatrix(Mtx44 matrixResult, Mtx44 matrix, f32 rScale);
-#pragma GLOBAL_ASM("asm/non_matchings/frame/frameScaleMatrix.s")
+bool frameScaleMatrix(Mtx44 matrixResult, Mtx44 matrix, f32 rScale) {
+    matrixResult[0][0] = matrix[0][0] * rScale;
+    matrixResult[0][1] = matrix[0][1] * rScale;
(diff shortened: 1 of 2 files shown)


src/emulator/frame.c
```diff
@@ -2701,14 +2888,94 @@ static inline bool frameCopyMatrix(Mtx44 matrixTarget, Mtx44 matrixSource) {
     return true;
 }
 
-bool frameScaleMatrix(Mtx44 matrixResult, Mtx44 matrix, f32 rScale);
-#pragma GLOBAL_ASM("asm/non_matchings/frame/frameScaleMatrix.s")
+bool frameScaleMatrix(Mtx44 matrixResult, Mtx44 matrix, f32 rScale) {
```
Yanis42, line 2891: any reason why they wouldn't use a loop here? (just wondering)

src/emulator/frame.c
```diff
@@ -2783,9 +3095,156 @@ static inline bool frameResetCache(Frame* pFrame) {
     return true;
 }
 
-#pragma GLOBAL_ASM("asm/non_matchings/frame/frameUpdateCache.s")
+static bool frameUpdateCache(Frame* pFrame) {
+    // s32 nCount;
+    // s32 nCountFree;
```
Yanis42, line 3100: is this intended?

Yanis42 (commented):
nice!

src/emulator/frame.c
cadmic: yeah, it's more variables in the debug info that I'm not sure how to use

src/emulator/frame.c
cadmic: if I had to guess, maybe not trusting the compiler to unroll the loops?

Yanis42 (approved):

PR #128, opened 2024-05-19
Declare non-static functions and add non-matching compatibility on ``mcardGCN.c``

Yanis42:
I was thinking it might be good for matching the functions, also I passed ``pMCard`` as a parameter for ``mcardUpdate``, based on the other functions I'd say it's an unused parameter but decomp will tell us for sure, let me know if I should revert this to ``void`` instead

also I think ``__anon_0x1A5F0`` is probably ``MemCardMessage`` but the usage confused me a bit so I didn't name it

### DIFF
--- a/src/emulator/mcardGCN.c
+++ b/src/emulator/mcardGCN.c
@@ -19,18 +19,13 @@ void* jtbl_800EA5A8[24] = {
 void* jtbl_800EA5A8[24] = {0};
 #endif
 
-#ifndef NON_MATCHING
 void* jtbl_800EA608[24] = {
     &lbl_80016E54, &lbl_80016E54, &lbl_80016DE8, &lbl_80016DC8, &lbl_80016E28, &lbl_80016DD8,
     &lbl_80016E08, &lbl_80016E54, &lbl_80016E54, &lbl_80016E54, &lbl_80016E54, &lbl_80016E54,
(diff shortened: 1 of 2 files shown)


src/emulator/mcardGCN.c
```diff
@@ -8,20 +8,29 @@ char D_800EA548[] =
 char D_800EA564[] = "Invalid Memory Card Command %d - Assuming Go To Game";
 char D_800EA59C[] = "mcardGCN.c";
 
+#ifndef NON_MATCHING
```
cadmic, line 11: if NON_MATCHING=1 aren't these jump tables still needed by some of the GLOBAL_ASM functions?

Yanis42: I just thought I'd do like the other places where we do this, is it a bad idea?

cadmic: If we do this now, then all the jump tables will get zeroed even though they're still used by the assembly versions which will definitely crash. It makes sense to do this if we replace the asm version with a C version, since then this jumptable is no longer used (and otherwise we'd get linker errors from `lbl_80016D74` etc not being defined)

Yanis42: I reverted the mcardUpdate parameter too, turns out it's really void according to the usage I'm seeing right now

cadmic: sorry I think you missed one

Yanis42: oops

PR #129, opened 2024-05-19
Match ``soundGCN.c``

Yanis42:
this is also fixing an issue with the decompme script, I added a type hint and two new regex, also I had to add something to fix float ordering, and ``soundMakeBuffer`` can probably be matched better but I think it's good enough for now


src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+        ((s16*)pSound->pBufferZero)[iData] = 0;
+    }
+
+    DCStoreRange(pSound->pBufferZero, pSound->nSizeZero);
+    return true;
+}
+
+bool soundPlayBuffer(Sound* pSound) {
+    void* pData;
+    s32 iBufferPlay; // iBuffer
```
Yanis42, line 164: small note: I decided to rename this variable to make it consistent with ``iBufferMake``, let me know if I should change this

src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+            nSize = pSound->anSizeBuffer[iBufferPlay];
+            pData = pSound->apBuffer[iBufferPlay];
+            pSound->iBufferPlay = (iBufferPlay + 1) % 16;
+        }
+        pSound->eMode = SPM_PLAY;
+    } else if ((*((volatile SoundPlayMode*)&pSound->eMode) == SPM_RAMPQUEUED) || (pSound->eMode == SPM_RAMPPLAYED)) {
+        //! TODO: fake match?
+        pData = pSound->pBufferZero;
+        nSize = pSound->nSizeZero;
+        pSound->eMode = SPM_RAMPPLAYED;
```
cadmic, line 185: hm what's fake about this?

src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+
+bool soundGetDMABuffer(Sound* pSound, u32* pnSize) {
+    *pnSize = AIGetDMABytesLeft();
+    return true;
+}
+
+bool soundSetBufferSize(Sound* pSound, s32 nSize) {
+    int iBuffer;
+
+    if (nSize % 32) {
```
cadmic, line 296: `nSize % 32 != 0` is clearer IMO

src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+
+    iBuffer = pSound->iBufferMake;
+    bPlay = 0;
+    nSize = pSound->nSndLen;
+    nSamples = ((nSize * 32000) + 16000) / pSound->nFrequency;
+    cmp = ((nSamples >> 2) & ~7);
+    temp_r6 = pSound->anSizeBuffer[iBuffer] = nSamples & (~0x1F);
+    curBufP = pSound->apBuffer[iBuffer];
+
+    for (vol = 0x10000, sample = 0, j = 0; j < cmp * 2; j += 2) {
```
cadmic, line 223: I think `vol = 0x10000` or `sample = 0` would make more sense before the loop

src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+    s32 nSize;
+    s32 samp;
+    s32 iBuffer;
+    s32 vol;
+    s32 bFlag;
+    s32 bPlay;
+    s32 volAdjust;
+
+    s32 temp_r6;
+    u32 cmp;
```
cadmic, line 213: I find `cmp` to be a particularly misleading name, I honestly think `temp_r3` is better if we don't want to think of a name now and it at least signals that this is probably fake to the reader

src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+    for (vol = 0x10000, sample = 0, j = 0; j < cmp * 2; j += 2) {
+        samp = sample >> 16;
+
+        if (j + 1 < (pSound->nSizePlay >> 1)) {
+            volAdjust = gVolumeCurve[vol >> 8];
+            curBufP[j + 1] = (volAdjust * ((s16*)pSound->pSrcData)[2 * samp + 0]) >> 16;
+            curBufP[j + 0] = (volAdjust * ((s16*)pSound->pSrcData)[2 * samp + 1]) >> 16;
+        }
+
+        vol = (u32)vol;
```
cadmic, line 232: could you leave a comment on this? otherwise it will probably confuse readers

src/emulator/soundGCN.c
cadmic: ehhh I kind of lean towards staying consistent with the debug data whenever possible. I'm not convinced that `iBufferMake` is real, for instance
```
if ((iBuffer = pSound->iBufferPlay) != pSound->iBufferMake) {
   ...
}
```
also matches

src/emulator/soundGCN.c
```diff
@@ -8,44 +13,444 @@ _XL_OBJECTTYPE gClassSound = {
...
+    if (!soundMakeBuffer(pSound)) {
+        return false;
+    }
+
+    return true;
+}
+
+bool soundSetDACRate(Sound* pSound, s32 nDacRate) {
+    pSound->nDacrate = nDacRate;
+    pSound->nFrequency = 0x02E6D354 / (nDacRate + 1);
```
cadmic, line 279: should be decimal, this is `VI_NTSC_CLOCK` from N64 rcp.h (found by googling this number):

```
/*
 * Possible video clocks (NTSC or PAL)
 */
#define VI_NTSC_CLOCK		48681812        /* Hz = 48.681812 MHz */
#define VI_PAL_CLOCK		49656530        /* Hz = 49.656530 MHz */
#define VI_MPAL_CLOCK		48628316        /* Hz = 48.628316 MHz */
```

src/emulator/soundGCN.c
Yanis42: the condition above, I wasn't sure where to place that comment ``else if ((*((volatile SoundPlayMode*)&pSound->eMode) == SPM_RAMPQUEUED) || (pSound->eMode == SPM_RAMPPLAYED))``

src/emulator/soundGCN.c
Yanis42: I made a macro in the file to make it clearer, let me know if I should move it somewhere else (I didn't know where to place it)

src/emulator/soundGCN.c
cadmic: I see, the current placement makes it seem like it's for the lines below, I'd either place it at the end of the line or the line before the problem

Yanis42: I moved it above the condition and I changed the comment to precise it's talking about the condition

PR #130, opened 2024-05-21
Match some functions in ``mcardGCN.c``

Yanis42:
that ``mcardPoll`` call from ``mcardCardErase`` seems to require ``!= 1``


src/emulator/mcardGCN.c
```diff
@@ -1,26 +1,38 @@
...
     &lbl_80016E54, &lbl_80016E54, &lbl_80016DE8, &lbl_80016DC8, &lbl_80016E28, &lbl_80016DD8,
     &lbl_80016E08, &lbl_80016E54, &lbl_80016E54, &lbl_80016E54, &lbl_80016E54, &lbl_80016E54,
     &lbl_80016E54, &lbl_80016E08, &lbl_80016E54, &lbl_80016E54, &lbl_80016DF8, &lbl_80016E54,
     &lbl_80016E44, &lbl_80016E18, &lbl_80016E18, &lbl_80016E54, &lbl_80016E54, &lbl_80016E34,
 };
+#else
+void* jtbl_800EA608[24] = {0};
+#endif
 
+#ifndef NON_MATCHING
```
cadmic, line 35: I think jtbl_800EA608 is for mcardMenu, which doesn't have a C version yet in NON_MATCHING mode

src/emulator/mcardGCN.c
```diff
@@ -113,23 +145,345 @@ static s32 bWrite2Card;
...
+}
+#endif
+
+bool mcardWrite(MemCard* pMCard, s32 address, s32 size, char* data) {
+    s32 i;
+    char testByte;
+
+    testByte = pMCard->file.game.buffer[0];
+    bWrite2Card = 1;
+    bNoWriteInCurrentFrame[currentIdx] = 0;
```
cadmic, line 419: should these both be bools?

src/emulator/mcardGCN.c
```diff
@@ -113,23 +145,345 @@ static s32 bWrite2Card;
...
+            CARDUnmount(pMCard->slot);
+        }
+    }
+
+    return true;
+}
+
+bool mcardGameRelease(MemCard* pMCard) {
+    if (pMCard->bufferCreated == 0) {
+        if (pMCard->file.game.buffer != NULL && !xlHeapFree(&pMCard->file.game.buffer)) {
```
cadmic, line 297: optional: I think these conditions are slightly more readable as

```
if (pMCard->file.game.buffer != NULL) {
    if (!xlHeapFree(&pMCard->file.game.buffer)) {
        return false;
    }
}
```
(ditto with `pMCard->file.game.writtenBlocks` and `!mcardUpdate()`)

PR #131, opened 2024-05-21
Match AmcExi2Stubs & odenotstub

Yanis42:
also adds splits for the 3 debugger files on ce-u, the ``AmcExiError`` enum is from pikmin2


src/debugger/AmcExi2Stubs.c
```diff
@@ -0,0 +1,17 @@
+#include "debugger.h"
+
+void EXI2_Init(void) {}
+
+void EXI2_EnableInterrupts(void) {}
+
+s32 EXI2_Poll(void) { return 0; }
```
cadmic, line 7: should `EXI2_Poll`, `AMC_IsStub`, and `Hu_IsStub` return bool?

cadmic (commented):
phew, these files looked difficult

src/debugger/AmcExi2Stubs.c
Yanis42: I was wondering the same thing

Yanis42:
probably my biggest challenge on this decomp so far

PR #135, opened 2024-05-22
Match GX files

Yanis42:
thanks pikmin2 & dolsdk2004 (sorry big pr)


include/macros.h
```diff
@@ -69,4 +69,11 @@ inline void padStack(void) { int pad = 0; }
 #define AT_ADDRESS(xyz)
 #endif
 
+#ifdef __MWERKS__
+#define rlwimi(a, b, c, d, e) __rlwimi((a), (b), (c), (d), (e))
+#else
+// using zero makes CC-Check happy
```
cadmic, line 75: I don't think a reader will know what "CC-Check" is tbh. What about declaring __rlwimi somewhere and use that everywhere instead of `rlwimi`? It can just stay undefined with gcc (while mwcc will magically replace it)
```
unsigned int __rlwimi(unsigned int rs, unsigned int is, unsigned int shift, unsigned int mask);
```

(I copied the signature from https://www.ibm.com/docs/en/openxl-c-and-cpp-aix/17.1.2?topic=rf-rldimi-builtin-ppc-rldimi-rlwimi-builtin-ppc-rlwimi but it might be missing an argument)

The __cntlzw intrinsic has a similar problem, and there's a few in math.h as well. Do other projects have a standard place to put these? If not, I propose `include/intrinsics.h` or something

src/dolphin/gx/GXFifo.c
```diff
@@ -0,0 +1,289 @@
...
+    GX_SET_CP_REG(2, gx->cpClr);
+}
+
+void __GXCleanGPFifo(void) {
+    bool interrupts;
+    GXFifoObjPriv tempObj; // 0x14
+    u32 i, j, k; // stack alloc
+    GXFifoObjPriv* gpFifo; // r31
+    GXFifoObjPriv* cpuFifo; // r30
+    void* tempPtr; // r29
```
cadmic, line 241: should we remove these comments?

src/dolphin/gx/GXInit.c
```diff
@@ -0,0 +1,544 @@
...
+    0x80000, 0x10000, 0xA0000, 0x30000, 0x88000, 0x50000, 0xA8000, 0x70000, 0x00000, 0x90000, 0x20000, 0xB0000,
+    0x40000, 0x90000, 0x60000, 0xB0000, 0x80000, 0x10000, 0xA0000, 0x30000, 0x80000, 0x50000, 0xA0000, 0x70000,
+};
+
+// forward declaring static reset function
+static bool __GXShutdown(bool);
+
+static OSResetFunctionInfo GXResetFuncInfo = {__GXShutdown, OS_RESET_PRIO_GX};
+
+#pragma peephole off
```
cadmic, line 59: maybe comment which function we know this is caused by?

src/dolphin/gx/GXInit.c
```diff
@@ -0,0 +1,544 @@
...
+
+static inline void EnableWriteGatherPipe(void) {
+    u32 hid2; // r31
```
cadmic, line 62: remove decomp comments?

src/dolphin/gx/GXTransform.c
```diff
@@ -0,0 +1,293 @@
+#include "dolphin/gx.h"
+#include "dolphin/mtx.h"
+
+inline void WriteProjPS(const register f32 src[6], register volatile void* dst) {
+    register f32 ps_0, ps_1, ps_2;
+
+#ifdef __MWERKS__ // clang-format off
+	asm {
```
cadmic, line 8: tabs -> spaces

include/macros.h
Yanis42: it seems these files are using 5 parameters in that function

cadmic: yeah `unsigned int __rlwimi(unsigned int a, unsigned int b, unsigned int c, unsigned int d, unsigned int e)` is probably fine

Yanis42: I had to remove the ``unsigned``s to make it build, no idea what's wrong with that but now it builds

PR #137, opened 2024-05-22
Mark Dolphin OS files as matched for ce-u

Yanis42:


cadmic:
thanks, want to check PPCArch too?

Yanis42:
oh yeah I didn't notice it was marked as nonmatching

PR #142, opened 2024-05-23
Match ansi_files.c and abort_exit.c (and related files)

cadmic:


src/libc/critical_regions.ppc_eabi.c
```diff
@@ -0,0 +1 @@
+void __kill_critical_regions(void) { return; }
```
Yanis42, line 1: I'm sure this was hard to match

Yanis42 (approved):
nice, I couldn't figure out the ``FILE`` struct when I looked at it yesterday

PR #143, opened 2024-05-23
Match CARD files

Yanis42:
thanks to pikmin2 for most files and mario kart double dash for ``CARDDelete.c``

ah also there was an unused file in configure.py so I removed it (CARDRename.c)


src/dolphin/card/CARDStat.c
```diff
@@ -0,0 +1,135 @@
+#include "dolphin/card.h"
+
+static void UpdateIconOffsets(CARDDir* entry, CARDStat* state) {
+    u32 offset;
+    bool iconTlut;
+    int i;
+
+    offset = entry->iconAddr;
+    if (offset == 0xffffffff) {
```
cadmic, line 9: 0xffffffff -> 0xFFFFFFFF here too?

src/dolphin/card/CARDUnlock.c
```diff
@@ -0,0 +1,386 @@
...
+#define SEC_BA(x) ((u8)(((x) >> 12) & 0x7f))
+
+inline int CARDRand() {
+    next = next * 1103515245 + 12345;
+    return (int)((u32)(next / 65536) % 32768);
+}
+
+/**
+ * @note Address: N/A
+ * @note Size: 0x8
```
cadmic: these `@note`s seem a bit out-of-place

Yanis42: ahh looks like I already removed it

src/dolphin/card/CARDStat.c
Yanis42: I hate lowercase hex, I need to figure out something to fix the other instances of it

PR #148, opened 2024-05-23
Match the remaining files of the Dolphin SDK

Yanis42:
thanks to double dash & pikmin2


src/dolphin/thp/THPDec.c
```diff
@@ -0,0 +1,2247 @@
...
+    __THPInfo->nMCU = (u16)((__THPInfo->file)[0] << 8 | (__THPInfo->file)[1]);
+    __THPInfo->file += 2;
+    __THPInfo->currMCU = __THPInfo->nMCU;
+}
+
+static inline void __THPGQRSetup(void) {
+    register u32 tmp1, tmp2;
+
+#ifdef __MWERKS__ // clang-format off
+	asm {
```
cadmic, line 425: no tabs pls

Yanis42: I hate that I miss those because clang has to be disabled there

PR #149, opened 2024-05-23
Fake weak math functions

cadmic:
The MSL C math library has some "weak" functions with C++ names that seem to get emitted after their first use, even if that first use is in an unused/stripped function. I tried doing this properly and gave up, so I just faked it in C instead.

Additionally, sqrt() is defined in math_inlines.c and used in e_pow.c even though it's inlined everywhere else. I'm not totally sure how this came to pass so I used some macro magic instead. The idea is that e_pow.c can do

```
#define SQRT_DECL WEAK
#include "math.h"
```

so it can use `sqrt()` without it getting defined or inlined.

I also reorganized math.h a bit (okay, a lot).


src/libc/e_pow.c
```diff
@@ -0,0 +1,284 @@
...
+    f64 y1, t1, t2, r, s, t, u, v, w;
+    int i, j, k, yisint, n;
+    int hx, hy, ix, iy;
+    u32 lx, ly;
+
+    hx = __HI(x);
+    lx = __LO(x);
+    hy = __HI(y);
+    ly = __LO(y);
+    ix = hx & 0x7fffffff;
```
Yanis42, line 54: there's lowercase hex in this file

Yanis42 (commented):
nice

src/libc/e_pow.c
cadmic: I can feel your rage

Yanis42 (approved):

PR #150, opened 2024-05-24
Match MQ-J

Yanis42:
this doesn't count SDK/libraries, it will be part of another PR (or PRs)

~~stuff I don't like/not satisfied with:~~
- ~~the ``INLINE`` and ``STATIC`` macros I added, on other versions these need to be changed so this implementation isn't great, we could define it in the files they're used instead~~
- ~~that switch in ``systemSetupGameRAM``, I hate this switch so much and it will get worse with other versions~~


src/emulator/simGCN.c
```diff
@@ -1899,9 +1926,20 @@ bool simulatorRumbleStop(s32 channel) {
...
+#define prevAllowResetSetting true
+bool simulatorTestReset(bool IPL, bool forceMenu, bool allowReset)
+#else
+bool simulatorTestReset(bool IPL, bool forceMenu, bool allowReset, bool usePreviousSettings)
+#endif
+{
     u32 bFlag;
     u32 nTick;
+
+#if VERSION != MQ_J
```
cadmic, line 1942: I think this would be slightly better if the cases were in the same order as the `#if` above. Also, maybe consider moving the `#define prevIPLSetting false` to here for better parallel structure?

src/emulator/simGCN.c
```diff
@@ -1972,10 +2013,12 @@ bool simulatorTestReset(bool IPL, bool forceMenu, bool allowReset, bool usePrevi
 }
 
 bool simulatorDrawMCardText(void) {
+#if VERSION != MQ_J
     if ((s32)(((TEXPalette*)gpErrorMessageBuffer)->versionNumber) == 0) {
         xlPostText("Invalid Message Image Data - Assuming SV09", "simGCN.c", 1623);
         simulatorPrepareMessage(S_M_CARD_SV09);
     }
+#endif
```
cadmic, line 2021: in general I think some more line breaks could help the reader group the `#if` and `#endif` together as they scan the code (e.g. blank line after the `#endif` here). I think would help with all these cases in simGCN.c and system.c. I think it's especially needed in places where a blank line is replaced with `#if/#endif` in the diff, I think it would be better to add `#if/#endif` without also removing the blank line

src/emulator/simGCN.c
```diff
@@ -340,23 +349,41 @@ bool simulatorDVDShowError(s32 nStatus, void* anData, s32 nSizeRead, u32 nOffset
...
+            if (gDVDResetToggle == 1 && (((u32)nStatus <= 3) || (((u32)nStatus - 7) <= 1) || (nStatus == 10))) {
+                if (!simulatorTestReset(false, false, true)) {
+                    return false;
+                }
+            } else if (!simulatorTestReset(true, false, true)) {
+                return false;
+            }
+        }
+#else
+    // clang-format off
```
cadmic, line 367: heh, I wonder if we can save this by keeping the `do { ... } while()` but changing it to `do { ... } while (false)` (even by setting continueToggle to false?)

src/emulator/system.c
```diff
@@ -179,6 +183,27 @@ static bool systemSetupGameRAM(System* pSystem) {
     if (romTestCode(pROM, "CZLJ") || romTestCode(pROM, "CZLE") || romTestCode(pROM, "NZSJ") ||
         romTestCode(pROM, "NZSE")) {
         switch (nCode) {
+#if VERSION == MQ_J
```
cadmic, line 186: hmm, maybe it's worth trying to figure the best way to write this for all 6 versions now, so we don't have to keep rewriting this. Maybe it really is best to write out 6 different cases here?

src/emulator/system.c
```diff
@@ -1487,6 +1553,12 @@ bool systemReset(System* pSystem) {
             return false;
         }
 
+#if VERSION == MQ_J
+        if (!libraryUpdate(SYSTEM_LIBRARY(pSystem))) {
+            return false;
+        }
+#else
+
```
cadmic, line 1561: nit: I think this looks a little strange with a blank line only on one side

src/emulator/system.c
cadmic: It might also help to figure out which ROMS/games these are, IIRC we figured out what the gnFlagZelda bits were already I think

src/emulator/simGCN.c
Yanis42: not sure I understood everything so let me know if what I did is fine (once I pushed)

src/emulator/system.c
Yanis42: it's way cleaner with the 6 cases written for each versions imo

PR #153, opened 2024-05-26
Match __mem.c and runtime.c

Yanis42:


src/emulator/mcardGCN.c
```diff
@@ -3,6 +3,7 @@
 #include "emulator/simGCN.h"
 #include "emulator/xlHeap.h"
 #include "emulator/xlPostGCN.h"
+#include "mem_funcs.h"
```
cadmic, line 6: nit: string.h is probably more "standard"

libc/mem_funcs.h
```diff
@@ -1,14 +1,21 @@
 #ifndef _MEM_FUNCS_H_
 #define _MEM_FUNCS_H_
 
+#include "stddef.h"
+
 #define __min_bytes_for_long_copy 32
 
+#pragma section code_type ".init"
```
cadmic, line 8: I've only just noticed, but why this pragma instead of `INIT`?

PR #154, opened 2024-05-31
Match more functions of ``mcardGCN.c``

Yanis42:


src/emulator/mcardGCN.c
```diff
@@ -115,9 +122,45 @@ static bool mcardWriteFileHeader(MemCard* pMCard);
...
+
+    pMCard->error = MC_E_NONE;
+    pMCard->slot = 0;
+    pMCard->wait = false;
+    pMCard->pPollFunction = NULL;
+    pMCard->writeToggle = true;
+
+    if (pMCard->writeBuffer != NULL && !xlHeapFree(&pMCard->writeBuffer)) {
+        return false;
+    }
```
cadmic, line 148: I think
```
if (pMCard->writeBuffer != NULL)
    if (!xlHeapFree(&pMCard->writeBuffer)) {
        return false;
    }
}
```
is a bit easier to read

src/emulator/mcardGCN.c
```diff
@@ -496,4 +1005,99 @@ bool mcardWrite(MemCard* pMCard, s32 address, s32 size, char* data) {
...
+    mcardOpenDuringGame(&mCard);
+
+    if (mCard.saveToggle == true) {
+        for (i = 0; i < (u32)(mCard.file.game.size + 8187) / 8188; i++) {
+            mCard.file.game.writtenBlocks[i] = 1;
+        }
+
+        prevIndex = 100;
+        counter = 0;
+        while (true) { // ?
```
cadmic, line 1028: why the `// ?`?

Yanis42: forgot to remove it lol, during decompilation I wasn't sure if that loop was really a while true so I added this to remind me this could cause issue if I'm getting stuck

PR #158, opened 2024-06-04
Match Dolphin SDK for MQ-J

Yanis42:
GX library will be part of another PR (I'm almost done with it)


src/dolphin/dsp/dsp.c
```diff
@@ -4,7 +4,16 @@
 #include "dolphin/types.h"
 
 bool __DSP_init_flag;
-char* __DSPVersion = "<< Dolphin SDK - DSP\trelease build: Apr 17 2003 12:34:16 (0x2301) >>";
+
+#if IS_MQ
+#define BUILD_DAY "Sep  5 2002"
```
cadmic, line 9: nit: BUILD_DAY -> BUILD_DATE
but why does only this file get this treatment and not the others? I don't care either way but it would be nice to be consistent

src/dolphin/dvd/dvd.c
```diff
@@ -1263,13 +1285,22 @@ bool DVDCheckDisk() {
...
+            }
+#else
             if (((coverReg >> 2) & 1) || (coverReg & 1)) {
                 result = false;
             } else if (ResumeFromHere != 0) {
                 result = false;
             } else {
                 result = true;
             }
+#endif
```
cadmic, line 1303: can also do something like this though idk if it's more readable
```
if (((coverReg >> 2) & 1) || (coverReg & 1)) {
    result = false;
} else {
#if IS_MQ
    result = true;
#else
    if (ResumeFromHere != 0) {
        result = false;
    } else {
        result = true;
    }
#endif
}
```

src/dolphin/exi/EXIBios.c
```diff
@@ -200,16 +199,24 @@ bool EXISync(s32 chan) {
             enabled = OSDisableInterrupts();
             if (exi->state & STATE_SELECTED) {
                 CompleteTransfer(chan);
-                if (__OSGetDIConfig() != 0xFF
-#if VERSION > MQ_E
-                    || ((OSGetConsoleType() & 0xF0000000) == OS_CONSOLE_TDEV)
-#endif
```
cadmic, line 206: FWIW I kinda liked this before, a bit clearer on what the differences are

configure.py
```diff
@@ -259,69 +259,69 @@ def MatchingFor(*versions):
     DolphinLib(
         "base",
         [
-            Object(MatchingFor("ce-j", "ce-u"), "dolphin/base/PPCArch.c"),
+            Object(MatchingFor("mq-j", "ce-j", "ce-u"), "dolphin/base/PPCArch.c"),
```
cadmic, line 262: can the `DOLPHIN_REV` define go away now?

src/dolphin/dsp/dsp.c
Yanis42: the only other place with the build date and time used several times is in OS.c, I just forgot to make the change since I already had OS files matched (just had to remove the ``DOLPHIN_REV`` macro)

src/dolphin/vi/vi.c
```diff
@@ -843,20 +847,24 @@ static void GetCurrentDisplayPosition(u32* hct, u32* vct) {
 }
 
 static inline u32 getCurrentHalfLine(void) {
+#if IS_MQ
     u32 hcount;
+    u32 vcount0;
     u32 vcount;
 
-#if VERSION <= MQ_E
-    u32 vcount0;
```
cadmic, line 850: could you move this local variable inside the `do { ... } while` so that there doesn't need to be an ifdef around hcount/vcount?

src/dolphin/os/OSReboot.c
```diff
@@ -29,15 +29,53 @@ extern u32 OS_RESET_CODE AT_ADDRESS(0x800030F0);
 extern u8 OS_REBOOT_BOOL AT_ADDRESS(0x800030E2); // unknown function, set to true by __OSReboot
 extern s32 __OSIsGcam;
 
+#if IS_MQ
+static volatile bool Prepared = false;
+#else
 static bool Prepared = false;
+#endif
```
cadmic, line 37: can this be volatile in both versions? It would be strange if they removed it

src/dolphin/card/CARDOpen.c
```diff
@@ -23,20 +23,42 @@ bool __CARDCompareFileName(CARDDir* entry, const char* fileName) {
     return false;
 }
 
+#if IS_MQ
+#define DISK_ID card->diskID
+#else
+#define DISK_ID diskID
+#endif
```
cadmic, line 30: I would combine this with the ifdef in the function (since otherwise `card->diskID` makes no sense outside of that function)

src/dolphin/demo/DEMOPuts.c
```diff
@@ -47,6 +47,10 @@ void DEMOLoadFont(enum _GXTexMapID texMap, enum _GXTexMtx texMtx, DMTexFlt texFl
 void DEMOSetupScrnSpc(s32 width, s32 height, f32 depth) {
     Mtx44 pMtx;
     Mtx mMtx;
+
+#if IS_MQ
+    C_MTXOrtho(pMtx, 0.0f, (f32)height, 0.0f, (f32)width, 0.0f, -depth);
+#else
```
cadmic, line 53: this is a bit simpler and matches both:
```
    Mtx44 pMtx;
    Mtx mMtx;
    f32 top = 0.0f;

#if IS_CE
    if (DEMOGetRenderModeObj()->field_rendering && !VIGetNextField()) {
        top = -0.667f;
    } else {
        top = 0.0f;
    }
#endif

    C_MTXOrtho(pMtx, top, (f32)height, 0.0f, (f32)width, 0.0f, -depth);
```

src/dolphin/os/__start.c
```diff
@@ -153,7 +153,7 @@ WEAK ASM void __start(void) {
 ASM void __init_registers(void) {
 #ifdef __MWERKS__ // clang-format off
     nofralloc
-    #if DOLPHIN_REV == 2003
+    #if IS_CE
```
cadmic, line 156: nit: unindent

src/dolphin/pad/Pad.c
```diff
@@ -703,7 +722,13 @@ bool __PADDisableRecalibration(bool disable) {
 
     enabled = OSDisableInterrupts();
     prev = (GameChoice & 0x40) ? true : false;
+
+#if IS_MQ
+    GameChoice &= 0xBF;
```
cadmic, line 727: `~0xC0`?

src/dolphin/vi/vi.c
Yanis42: if vcount0 is declared after vcount it's messing up the regalloc, also it's used in that do while's condition so declaring inside it wouldn't work anyway I think

src/dolphin/os/OSReboot.c
Yanis42: seems like it works

PR #159, opened 2024-06-05
Add support for windows

Yanis42:
on Linux running the compile.sh script with ``sh`` fails, on windows using ``sh.exe`` works, using ``bash`` works on Linux but ``bash.exe`` doesn't work on Windows, hence that ``else`` I added in ``project.py``

also I'm not a bash expert so let me know if anything needs to be changed in ``compile.sh``

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,17 @@ brew install git ninja python3
 brew install --cask --no-quarantine gcenx/wine/wine-crossover
 ```
 
+#### Windows
+
+You will need the following dependencies:
+- [ninja.exe](https://github.com/ninja-build/ninja/releases/latest)
+- Python (make sure to add it to your PATH during the installation)
+- [Git for Windows](https://www.git-scm.com/downloads)
+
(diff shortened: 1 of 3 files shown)


tools/asm_processor/compile.sh
```diff
@@ -2,6 +2,14 @@
 
 set -euo pipefail
 
+# bash scripts are executed with sh.exe, installed with Git for Windows, which is using MSYS
+OS=$(uname -o)
+PYTHON="python3"
+if [ "$OS" == "Msys" ]
+then
```
cadmic, line 9: personally I like this style better:
```
if [ "$OS" == "Msys" ]; then
    ....
end
```

README.md
```diff
@@ -44,6 +44,19 @@ brew install git ninja python3
 brew install --cask --no-quarantine gcenx/wine/wine-crossover
 ```
 
+#### Windows
+
+If you want to use WSL, follow the steps for Ubuntu/Debian above.
```
cadmic, line 49: the wording here makes it sound like the rest of the instructions is what you need to do if you want to use WSL. Maybe call this section `Windows (native, not WSL)` or something? not sure.

PR #160, opened 2024-06-06
Match GX library on MQ

Yanis42:
draft for now, GXLight is only missing two functions and they share the same issue so I thought it would be ok to open the PR now

not entirely satisfied with the code in these files but it works for both versions 🤷


Yanis42:
I merged main and added mq-u, the only file left is ``GXLight`` but I don't feel motivated enough so I'm marking this as ready to review

include/dolphin/gx/GXData.h
```diff
@@ -126,8 +126,14 @@ typedef struct _GXData {
 
     // Texture regions
     /* 0x208 */ GXTexRegion TexRegions0[GX_MAX_TEXMAP];
+#if IS_MQ
+    /* 0x288 */ GXTexRegion TexRegionsCI[4];
+    /* 0x2C8 */ u32 nextTexRgn;
+    /* 0x2CC */ u32 nextTexRgnCI;
```
cadmic, line 132: why put the offsets for some fields but not others?

It might be confusing to mix offsets anyway, should we put them only for CE?

Yanis42: this struct is weird because of the version differences, I had no idea how to handle it properly

cadmic: I can think of two options:
* Only label offsets for CE
* Comment two of offsets for every field, one for MQ and one for CE

Yanis42: I was thinking of a 3rd option: have two separate structs one for each version, there's only two versions so hopefully it's not that bad

cadmic: Hmm this is a pretty big struct, personally I'd like to avoid that

include/dolphin/gx/GXData.h
```diff
@@ -55,133 +55,137 @@ typedef enum _CPClear {
...
+    /* 0x4F0 0x5A8 */ GXBool inDispList;
+    /* 0x4F1 0x5A9 */ GXBool dlSaveContext;
 #if IS_CE
-    /* 0x5AA */ GXBool abtWaitPECopy;
+    /*  N/A  0x5AA */ GXBool abtWaitPECopy;
 #endif
-    /* 0x5AB */ u8 dirtyVAT;
-    /* 0x5AC */ u32 dirtyState;
+    /* 0x4F2 0x5AB */ u8 dirtyVAT;
+    /* 0x4F3 0x5AC */ u32 dirtyState;
```
cadmic, line 188: Are you sure 0x4F3 is correct? It seems like `dirtyState` should be aligned to 4 bytes

Yanis42: I was thinking it's odd too

PR #165, opened 2024-07-28
mCardGCN work

Prakxo:


src/emulator/mcardGCN.c
```diff
@@ -73,25 +73,369 @@ static s32 bWrite2Card;
...
+            return false;
+        case -13:
+            pMCard->error = MC_E_ENCODING;
+            return false;
+        case -14:
+            pMCard->error = MC_E_CANCELED;
+            return false;
+        case -128:
+            pMCard->error = MC_E_FATAL;
+            return false;
```
Yanis42, line 127:
```suggestion
        case CARD_RESULT_READY:
            pMCard->error = MC_E_NONE;
            return true;
        case CARD_RESULT_BUSY:
            pMCard->error = MC_E_BUSY;
            return false;
        case CARD_RESULT_WRONGDEVICE:
            pMCard->error = MC_E_WRONGDEVICE;
            return false;
        case CARD_RESULT_NOCARD:
            pMCard->error = MC_E_NOCARD;
            pMCard->isBroken = 0;
            return false;
        case CARD_RESULT_NOFILE:
            pMCard->error = MC_E_NOFILE;
            return false;
        case CARD_RESULT_IOERROR:
            pMCard->error = MC_E_IOERROR;
            pMCard->isBroken = 1;
            return false;
        case CARD_RESULT_BROKEN:
            pMCard->error = MC_E_BROKEN;
            return false;
        case CARD_RESULT_EXIST:
            pMCard->error = MC_E_EXIST;
            return false;
        case CARD_RESULT_NOENT:
            pMCard->error = MC_E_NOENT;
            return false;
        case CARD_RESULT_INSSPACE:
            pMCard->error = MC_E_INSSPACE;
            return false;
        case CARD_RESULT_NOPERM:
            pMCard->error = MC_E_NOPERM;
            return false;
        case CARD_RESULT_LIMIT:
            pMCard->error = MC_E_LIMIT;
            return false;
        case CARD_RESULT_NAMETOOLONG:
            pMCard->error = MC_E_NAMETOOLONG;
            return false;
        case CARD_RESULT_ENCODING:
            pMCard->error = MC_E_ENCODING;
            return false;
        case CARD_RESULT_CANCELED:
            pMCard->error = MC_E_CANCELED;
            return false;
        case CARD_RESULT_FATAL_ERROR:
            pMCard->error = MC_E_FATAL;
            return false;
```
add ``#include "dolphin/card.h"`` if these defines aren't defined

src/emulator/mcardGCN.c
```diff
@@ -73,25 +73,369 @@ static s32 bWrite2Card;
...
+}
+
+static bool mcardCalculateChecksumFileBlock1(MemCard* pMCard, s32* checksum) {
+    s32 i;
+
+    if (mCard.saveToggle == true) {
+        *checksum = 0;
+
+        for (i = 0; i != 0x800; i++) {
+            if (i != 10U) {
```
Yanis42, line 158: you should remove the ``U`` there (and in the other places too, if it's matching)

src/emulator/mcardGCN.c
```diff
@@ -73,25 +73,369 @@ static s32 bWrite2Card;
...
+        memcpy(time, buf, 0x28);
+    }
+
+    return true;
+}
+
+static bool mcardReplaceFileBlock(MemCard* pMCard, s32 index) {
+    s32 checksum1;
+    s32 checksum2;
+    char buffer[8192];
```
cadmic, line 264: nit: I feel like this should be 0x2000 to be consistent with the args below

src/emulator/mcardGCN.c
```diff
@@ -73,25 +73,369 @@ static s32 bWrite2Card;
...
+
+    if (checksum1 != checksum2) {
+        pMCard->error = MC_E_CHECKSUM;
+        memcpy(pMCard->writeBuffer, buffer, 0x2000);
+        return false;
+    }
+
+    simulatorPrepareMessage(S_M_CARD_SV09);
+
+    if (UnkInlinemCardReplaceFileBlock2(pMCard, index << 13) == false) {
```
cadmic, line 289: does `if (!UnkInlinemCardReplaceFileBlock2(...))`? work? I think it should be preferred to `== false` everywhere

src/emulator/mcardGCN.c
```diff
@@ -73,25 +73,369 @@ static s32 bWrite2Card;
...
+    return true;
+}
+
+static bool mcardVerifyChecksumFileHeader(MemCard* pMCard) {
+    char* buffer;
+
+    if (UnkInlinemCardVerifyChecksumFileHeader(pMCard) == false) {
+        return false;
+    }
+    if (xlHeapTake(&buffer, 0x30006000) == 0) {
```
Yanis42, line 400:
```suggestion
    if (!xlHeapTake(&buffer, 0x6000 | 0x30000000)) {
```
hopefully the ``== 0`` change doesn't break the match (also do that for other functions like this too)

cadmic, line 400: !xlHeapTake etc.

src/emulator/mcardGCN.c
```diff
@@ -73,25 +73,369 @@ static s32 bWrite2Card;
...
+static bool mcardCalculateChecksumFileBlock2(MemCard* pMCard, s32* checksum) {
+    s32 i;
+
+    if (mCard.saveToggle == true) {
+        *checksum = 0;
 
-#pragma GLOBAL_ASM("asm/non_matchings/mcardGCN/mcardReplaceFileBlock.s")
+        for (i = 0; i != 0x800; i++) {
+            if (i != 0x41BU) {
+                *checksum += *((s32*)pMCard->writeBuffer + i);
```
cadmic, line 179: hmm, I think this would be clearer as `((s32*)pMCard->writeBuffer)[i]`, but it doesn't work everywhere. Later for `checksum != *(s32*)(pMCard->writeBuffer + 0x28)` it's kinda nice to see the byte offset. Maybe change this only in mcardCalculateChecksumFileBlock1 and mcardCalculateChecksumFileBlock2 because here it's clear it's iterating over the writeBuffer one word at a time? What do you think?

src/emulator/mcardGCN.c
Prakxo: Needs to match, they're cmplwi

src/emulator/mcardGCN.c
Prakxo: Seems okay by me

src/emulator/mcardGCN.c
Yanis42: it wouldn't match the debug data we have but does it work if the variable is turned to be unsigned instead? 🤔

cadmic: If it's too ugly, we could hide the U in a constant now or in the future (here 10U * 4 is reused as offset 0x28 below, and 0x41BU * 4 is reused below as offset 0x106C)

Yanis42: yeah I guess we can clean it up later 👍

Yanis42 (approved):

PR #166, opened 2024-07-28
Add ``ninja all``

Yanis42:


tools/project.py
```diff
@@ -526,6 +526,14 @@ def generate_build_ninja(
     )
     n.newline()
 
+    n.comment("Build All Versions")
```
cadmic, line 529: I think this should go near where `Default rule` and `split-all` are defined, this section is more for defining build command lines

Speaking of which, I think `split-all` should depend on only the versions in `build_configs` too. Could you fix that while you're at it? Sorry for making you clean up my mess

PR #167, opened 2024-07-29
Match some RSP functions

Yanis42:
and fixed the context generated from ``decompme.py``

matched:
- ``rspParseJPEG_EncodeZ``
- ``rspParseJPEG_DecodeZ``
- ``rspFillObjSprite``
- ``rspFillObjBgScale``
- ``rspFillObjBg``
- ``rspSetImage``
- ``tmemLoad_B``
- ``tmemLoad_A``

### DIFF
--- a/include/emulator/rsp.h
+++ b/include/emulator/rsp.h
@@ -309,12 +309,36 @@ typedef struct __anon_0x5F05A {
 } __anon_0x5F05A; // size = 0x28
 
 typedef union __anon_0x5F2FB {
-    /* 0x0 */ struct __anon_0x5ED4F b;
-    /* 0x0 */ struct __anon_0x5F05A s;
-    /* 0x0 */ s64 force_structure_alignment;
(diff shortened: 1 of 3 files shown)


include/emulator/rsp.h
```diff
@@ -309,12 +309,36 @@ typedef struct __anon_0x5F05A {
 } __anon_0x5F05A; // size = 0x28
 
 typedef union __anon_0x5F2FB {
-    /* 0x0 */ struct __anon_0x5ED4F b;
-    /* 0x0 */ struct __anon_0x5F05A s;
-    /* 0x0 */ s64 force_structure_alignment;
+    /* 0x0 */ __anon_0x5ED4F b;
+    /* 0x28 */ __anon_0x5F05A s;
+    /* 0x50 */ s64 force_structure_alignment;
```
cadmic, line 314: This is a union so all of its members will be at offset 0. Although maybe these comments are not useful at all

PR #168, opened 2024-08-04
Fixed ninja package name for Linux

Yanis42:
noticed this earlier, idk if the name is correct for macOS


cadmic:
seems correct for mac

PR #171, opened 2024-08-27
Remove InsertBraces from .clang-format (for old clang-format versions)

cadmic:
I don't think it adds too much so maybe it's not worth making users upgrade for this (some projects, like oot, require an old clang-format version). Lmk if you disagree though


Yanis42:
wasn't this added when we started working on libraries? it's the thing to add ``{}``s when it's missing right? it can go away I think

and if we really need it we can add it back on our local forks and remove it again before pushing

cadmic:
yeah, I always add the braces anyway when writing code so I wouldn't personally miss it

PR #173, opened 2024-09-04
CI fixes

cadmic:
Needs local windows testing

PR #176, opened 2024-09-05
Fix Windows build

cadmic:
Turns out #173 does not in fact work on Windows


Yanis42 (approved):
seems to work, this time I hope it's real and not another Windows prank

PR #179, opened 2024-09-18
Match the rest of mcardGCN.c

cadmic:
Sorry for the large PR, a lot of the code seems pretty repetitive. I'll try to do cleanups in a follow-up PR, namely:
* give names for constants
* use `true/false` consistently for boolean variables (like `saveToggle`)
* make a macro for `(u32)(x + 8187) / 8188`


src/emulator/mcardGCN.c
```diff
@@ -427,99 +459,921 @@ static inline bool mcardWriteAnywhereNoTime(MemCard* pMCard, s32 offset, u32 siz
...
+    if (!mcardTimeCheck(pMCard)) {
+        mCard.writeToggle = true;
+        return false;
+    }
+    if (!mcardReadAnywhere(pMCard, 0x2000, 0x2000, pMCard->writeBuffer)) {
+        mCard.writeToggle = true;
+        return false;
+    }
+
+    DCInvalidateRange(pMCard->writeBuffer, 0x2000U);
```
Yanis42, line 689: I assume the ``U``s are mandatory for matching?

src/emulator/mcardGCN.c
```diff
@@ -427,99 +459,921 @@ static inline bool mcardWriteAnywhereNoTime(MemCard* pMCard, s32 offset, u32 siz
...
+            }
+            if (OSGetSoundMode() == OS_SOUND_MODE_MONO) {
+                pMCard->file.game.buffer[0] &= 0xFC;
+                pMCard->file.game.buffer[0] |= 1;
+            } else if (OSGetSoundMode() == OS_SOUND_MODE_STEREO) {
+                if ((pMCard->file.game.buffer[0] & 0xF) == 1) {
+                    pMCard->file.game.buffer[0] &= 0xFC;
+                }
+            }
+            *pMCard->file.game.writtenBlocks = 1;
```
Yanis42, line 1124: this part looks familiar (from ``mcardWrite``), maybe it's an inline?

Yanis42 (approved):
I think a lot of those values are probably macros from the SDK
otherwise I think what I commented here are things that are planned for cleanup, nice work!

src/emulator/mcardGCN.c
Yanis42: actually it's not exactly the same but it's very similar

src/emulator/mcardGCN.c
cadmic: it isn't in this case, but I didn't bother since I'll give it a name anyway soon

src/emulator/mcardGCN.c
cadmic: yeah it is very similar. I looked through the erased functions and I'm not sure what it could be

PR #181, opened 2024-09-19
Use bool everywhere in mcard

cadmic:
Part 1 of mcard cleanups. For things named `toggle` I'm not sure whether to use `true/false` or `0/1` (like an enum, I guess), since they do `toggle == 1` a lot. Then again, the also do `someFunctionReturningABool() == true` a lot too so maybe that's just the style in this file.

### DIFF
--- a/include/emulator/mcardGCN.h
+++ b/include/emulator/mcardGCN.h
@@ -3,7 +3,6 @@
 
 #include "dolphin.h"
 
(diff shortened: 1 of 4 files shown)


Yanis42 (approved):
I was confused by ``mCard.file.game.writtenBlocks`` but it must be an array to know which blocks are written or something like that so it makes sense it's a bool too

PR #182, opened 2024-09-19
Document memory card format, add macros for offsets + sizes

cadmic:
I considered computing the sizes of everything (e.g. `sizeof(OSCalendarTime)` instead of 0x28) but it got weird since the size of the game names section is wrong in the format.

### DIFF
--- a/src/emulator/mcardGCN.c
+++ b/src/emulator/mcardGCN.c
@@ -1,3 +1,32 @@
+/**
+ * @file mcardGCN.c
+ *
+ * The memory card file can hold data for up to 16 games at once (although in practice at most 2 games are used,
+ * for Master Quest). The file starts with a 3 block header, where the first 2 blocks contain the following data:
+ *
+ *   offset   size    description
+ *   -----------------------------
+ *   +0x0000  0x0028  timestamp, as an OSCalendarTime
+ *   +0x0028  0x0004  checksum of the first block
+ *   +0x002C  0x0020  comment string
+ *   +0x004C  0x0020  "date string", always "ゼルダの伝説：時のオカリナ" ("The Legend of Zelda: Ocarina of Time")
+ *   +0x006C  0x1800  banner TPL data
+ *   +0x186C  0x1800  icon TPL data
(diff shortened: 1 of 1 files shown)


src/emulator/mcardGCN.c
```diff
@@ -7,9 +7,75 @@
...
+// block, the block data is first written to +0x4000 and read back to verify the card is working correctly.
+//
+// After the header, the file contains a sequence of games, where each game has a sequence of game data blocks
+// followed by 1 extra block. Each game data block contains a 4-byte checksum followed by 0x2000 - 4 = BLOCK_DATA_SIZE
+// bytes of data. The extra block is used as temporary space when writing similar to extra the header block above.
+//
+// Checksums are a simple sum of all 32-bit words in the block, with the exception of the checksum field itself.
+// If the checksum is 0, it is replaced with 1 to avoid having a valid block with all 0s.
+
+#define BLOCK_SIZE 0x2000
```
Yanis42, line 36: I think this should have a comment or something to tell that it's synced with the value from `CARD_SYSTEM_BLOCK_SIZE`

src/emulator/mcardGCN.c
```diff
@@ -7,9 +7,75 @@
 #include "stdio.h"
 #include "string.h"
 
-#define ALIGN_BUFFER(buffer) (char*)((buffer) + (0x20 - (s32)(buffer) % 32) % 32)
-
-static char gMCardCardWorkArea[40960];
+// The memory card file can hold data for up to 16 games at once (although in practice at most 2 games are used,
```
Yanis42, line 10: is it worth to do like
```c
/**
  * @file mcardGCN.c
  *
  * The memory card file can hold ...
  * 
  */ 
```
?

Yanis42 (commented):
nice work!

src/emulator/mcardGCN.c
cadmic: I don't really see the point if we don't use Doxygen, but I guess the functions docs are consistent with the oot repo so might as well do it for file comments too

Yanis42 (approved):

PR #184, opened 2024-09-20
Allow building without a default version (ce-j)

cadmic:
Currently if try to set up only mq-j, you can't, because project.py requires that `default_version` exists even though nothing relies on this (the default target for ninja is only generated if `default_version is not None`).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -121,8 +121,15 @@ config.versions = [
     for version in ALL_VERSIONS
     if (Path("orig") / version / "main.dol").exists()
 ]
+
+if not config.versions:
+    sys.exit("Error: no main.dol found for any version")
+
 if "ce-j" in config.versions:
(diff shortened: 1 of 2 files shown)


Yanis42:
it seems to work but the output when using ``ninja`` alone is a bit odd compared to ``ninja mq-j`` https://pastebin.com/y6L03Xqu

cadmic:
hmm I guess by default ninja tries to build literally everything if we don't set a default command. What should the default be in this case? We could also change it to `all` instead of trying to make it be `ce-j`

Yanis42:
I think the default should be ce-j and if it's not found the earlier version available

cadmic:
Alright, that seems reasonable

Yanis42 (approved):

PR #187, opened 2024-09-21
Add missing splits for MQ-E and CE-E

Yanis42:
This provides splits for the SDK and other libraries for MQ-E and CE-E, also I fixed a small mistake on the dtors splits for CE-J and CE-U (also I named most functions for MQ-E)

The only differences in the libs are ``OSGetLanguage`` and ``strcat``, which aren't present on the other versions


libc/string.h
```diff
@@ -10,6 +10,7 @@ size_t strlen(const char* s);
 int strcmp(const char* s1, const char* s2);
 int strncmp(const char* s1, const char* s2, size_t n);
 char* strchr(const char* str, int chr);
+char* strcat(char* dest, const char* src);
```
cadmic, line 13: nit: I'd order this just above `strncat` (or instead of `strncat`? I'm don't see anything that defines is)

cadmic (approved):

PR #188, opened 2024-09-21
Match some RSP audio functions

cadmic:
Mainly `rspParseABI[1-4]` and some of its helper functions. I mostly gave up on documenting because I'm not sure which ABI corresponds to which games (although I think ABI1 is sm64 and ABI2 is oot at least)

### DIFF
--- a/src/emulator/rsp.c
+++ b/src/emulator/rsp.c
@@ -113,19 +113,29 @@ void* jtbl_800EE568[27] = {
     &lbl_800810A4, &lbl_800810F0, &lbl_8008111C, &lbl_8008112C, &lbl_8008118C, &lbl_8008113C,
 };
 
+#ifndef NON_MATCHING
+// rspParseABI3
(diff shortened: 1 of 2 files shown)


src/emulator/rsp.c
```diff
@@ -105,37 +106,57 @@ void* jtbl_800EE4EC[31] = {
...
 };
+#else
+void* jtbl_800EE690[16] = {0};
+#endif
 
-static s32 nFirstTime_2148 = 0x00000001;
-static s32 nFirstTime_2648 = 0x00000001;
-static s32 nFirstTime_2757 = 0x00000001;
-static s32 nFirstTime_2796 = 0x00000001;
+static bool nFirstTime_2148 = true;
```
Yanis42, line 156: maybe add ``//! TODO: move to the functions`` or something like that, it seems to be static variables inside functions

Yanis42 (approved):
nice work!

PR #189, opened 2024-09-26
Match mcardGCN.c for mq-j, mq-u, ce-u and ce-e

Yanis002:
this doesn't do mq-e because I don't feel motivated enough atm but feel free to work on it, some parts aren't very pretty (``mcardWrite`` for instance) but I'm out of ideas, also should I add the accurate struct offsets? it's only a difference of 0x4 (``mCard->isBroken`` doesn't exist on mq-j)


cadmic:
I might try to play with making things look a little prettier, I'll try not to take too long though

Yanis002:
nice!

PR #190, opened 2024-10-01
Fix permuter tool paths

cadmic:
Stuff has moved around since the dtk migration


Yanis42 (approved):
works well but Windows users need to use WSL, which is fine for me as I'm already doing this anyway

PR #191, opened 2024-10-01
Match some jpeg functions

cadmic:
for-loops for days

### DIFF
--- a/src/emulator/rsp.c
+++ b/src/emulator/rsp.c
@@ -2409,22 +2409,462 @@ static bool rspInitAudioDMEM4(Rsp* pRSP) {
 
 #pragma GLOBAL_ASM("asm/non_matchings/rsp/rspParseABI4.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/rsp/rspCreateJPEGArrays.s")
+static bool rspCreateJPEGArrays(Rsp* pRSP) {
+    pRSP->Coeff = (s32*)pRSP->pDMEM;
+    pRSP->Coeff[0] = 0x1000;
+    pRSP->Coeff[1] = 0x1000;
+    pRSP->Coeff[2] = 0x1000;
+    pRSP->Coeff[3] = 0x1000;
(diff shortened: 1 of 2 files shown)


src/emulator/rsp.c
```diff
@@ -2433,25 +2873,484 @@ static bool rspParseJPEG_Encode(Rsp* pRSP, RspTask* pTask);
...
+    s32 g;
+    s32 b;
+    s32 y;
+    s32 u;
+    s32 v;
+
+    for (i = 0; i < 8; i++) {
+        for (j = 0; j < 8; j++) {
+            rspLoadColorBufferZ(pRSP, &r, &g, &b, imgBuf, i * 16 + j);
+            y = g + (b << 16) / 116195 + (r << 16) / 91914;
```
Yanis42, line 3320: out of curiosity, why 116195 and 91914?

Yanis42 (approved):
nice work!

src/emulator/rsp.c
cadmic: Using YUV (or more accurately Y'CbCr) is for https://en.wikipedia.org/wiki/Chroma_subsampling and is part of the JPEG standard. Basically, the human eye is more sensitive to changes in brightness than changes in color, so in the compressed file the brightness information (Y) is kept at a higher resolution than the color information (U and V). However you don't just want to do Y = R + G + B because the eye reacts differently to different colors (it's much more sensitive to green for example), so it uses some other mix where the constants were determined experimentally.

These particular constants match the ones in `rspUndoRecon420Z`, for example:

b = y + ((u * 0xC5E3) >> 16) + u
   = y + ((u * 0x1C5E3) >> 16)
   = y + ((u * 116195) >> 16)
   ≈ y + 1.773 u

which is close to the 1.772 at https://en.wikipedia.org/wiki/YCbCr#JPEG_conversion. For `rspRecon420Z` they seems to have tried to reverse that logic, using division instead of multiplication here. But the transformation seems totally incorrect.

PR #193, opened 2024-10-02
Match some S2DEX functions

cadmic:
The structs here are copy-pasted from N64 gs2dex.h so I gave them the same name. And a couple functions are basically copy-pastes from libultra us2dex_emu.c with some tweaks

### DIFF
--- a/include/emulator/rsp.h
+++ b/include/emulator/rsp.h
@@ -362,7 +362,6 @@ typedef struct uObjTxtrTile_t {
     /* 0x14 */ u32 mask;
 } uObjTxtrTile_t; // size = 0x18
 
-// uObjTxtrTLUT_t
 typedef struct uObjTxtrTLUT_t {
     /* 0x00 */ u32 type;
     /* 0x04 */ u32 image;
--- a/include/emulator/rsp.h
+++ b/include/emulator/rsp.h
@@ -268,8 +268,7 @@ typedef struct Rsp {
     /* 0x39C8 */ int* dctBuf;
 } Rsp; // size = 0x39CC
 
-// uObjBg_t
-typedef struct __anon_0x5ED4F {
+typedef struct uObjBg_t {
     /* 0x00 */ u16 imageX;
     /* 0x02 */ u16 imageW;
     /* 0x04 */ s16 frameX;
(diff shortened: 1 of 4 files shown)


include/emulator/rsp.h
```diff
@@ -329,16 +333,57 @@ typedef struct __anon_0x5F429 {
...
+    /* 0x04 */ u32 image;
+    /* 0x08 */ u16 tmem;
+    /* 0x0A */ u16 twidth;
+    /* 0x0C */ u16 theight;
+    /* 0x0E */ u16 sid;
+    /* 0x10 */ u32 flag;
+    /* 0x14 */ u32 mask;
+} uObjTxtrTile_t; // size = 0x18
+
+// uObjTxtrTLUT_t
```
Yanis42, line 365: I think this isn't necessary right?

src/emulator/rsp.c
```diff
@@ -158,8 +158,20 @@ static bool nFirstTime_2648 = true;
 static bool nFirstTime_2757 = true;
 static bool nFirstTime_2796 = true;
 
+static s32 counter;
+static u16 scissorX0;
+static u16 scissorY0;
 static u16 scissorX1 = 0x500;
 static u16 scissorY1 = 0x3C0;
```
Yanis42, line 165: these are 1280 and 960, it looks like definition values to me (so it should probably be decimal instead of hex)

src/emulator/rsp.c
```diff
@@ -4936,26 +4964,707 @@ static bool tmemLoad_A(Frame* pFrame, Rsp* pRSP, s32 imagePtr, s16 loadLines, s1
...
+                } else {
+                    imagePtr = imageTop - imageRemain * imageSrcWsize + imagePtrX0;
+                    imageRemain += tmemSrcLines;
+                }
+            } else {
+                Rectangle primitive;
+                s16 nS, nT;
+                s16 framePtrY1;
+
+                frameSliceCount &= 1023;
```
Yanis42, line 5307: shouldn't this value be in hex? though it seems the original function have them in decimal too 🤔

Yanis42 (commented):
nice!

src/emulator/rsp.c
cadmic: Yeah, changed to use N64_FRAME_WIDTH/N64_FRAME_HEIGHT like the original

src/emulator/rsp.c
cadmic: Yeah I kept it from the original version (same with declaring multiple variables on the same line)

Yanis42 (approved):

PR #196, opened 2024-10-14
Update badges to show the matched% instead of the linked%

Yanis42:


encounter:
FYI, you can use .svg which will be a little nicer for high DPI screens. I used .png when posting in discord for the preview

cadmic (approved):

PR #198, opened 2024-10-21
Match cpuGetPPC

cadmic:
This actually wasn't too bad, mostly just tedious. Unfortunately compiling cpu.c uses 1.5GB of RAM and takes about 1m20s to compile on a fast computer, and about 8m on my laptop.

I wasn't able to easily match data ordering so I'll leave that for a future PR.


Yanis42 (approved):
congratulations on matching this one 🎉

PR #199, opened 2024-10-24
Match and link cpu.c

cadmic:
I created _cpuGCN.c and _cpuDecodePPC.c, based on debug info and oot-vc function order, since apparently including other source files changes function ordering but not data ordering. I also fixed a few things that weren't actually matching.

Some things I'm unsure about:
* There's some unused strings which I think are part of the (stripped) `treePrint`, so I created a dummy function for these, but I forgot if there's another way to do it
* `cpuCountTLB` seems like it should return a boolean but I can't get that to match while using it in `cpuExecuteOpcode`. Maybe there's a different inline here?


docs/c_cpp_properties.json
```diff
@@ -8,6 +8,7 @@
                 "${workspaceFolder}/src",
                 "${workspaceFolder}/include",
                 "${workspaceFolder}/libc",
+                "${workspaceFolder}/src",
```
Yanis42, line 11: this is already in the includes (same thing for the json of the vscode.md file)

Yanis42 (approved):
congrats on this cpu work, it's really nice to see all the cleanup it brings

docs/c_cpp_properties.json
cadmic: huh thanks

PR #202, opened 2025-04-12
Match frame.c data

cadmic:
Some weirdness that could be better:
* I put `gaszNameColor` an friends in the middle of _frameGCN.c since it must come after `sRemapI`
* I put `anRenderModeDatabaseCycle2` at the end of frame.c, since it seems it can't be in _frameGCN.c (but must be after the `#include "emulator/_frameGCN.c")


configure.py
```diff
@@ -274,7 +274,7 @@ def LinkedFor(*versions):
             Object(LinkedFor("ce-j"), "emulator/mcardGCN.c"),
             Object(Linked, "emulator/codeGCN.c"),
             Object(Linked, "emulator/soundGCN.c"),
-            Object(LinkedFor("ce-j"), "emulator/frame.c", asm_processor=True),
+            Object(LinkedFor("ce-j"), "emulator/frame.c"),
```
Yanis002, line 277:
```suggestion
            Object(LinkedFor("mq-u", "mq-e", "ce-j", "ce-u", "ce-e"), "emulator/frame.c"),
```
I tried and this seems to match for those, also you might need to run `dol-apply`

Yanis002 (approved):
the data organization looks odd but it's fine imo, we can always go back to this later if it's really necessary

one of the biggest chunks of the emulator finally matched entirely! congrats :tada:

configure.py
Yanis002: I think I got baited by my build folder, I'm recompiling just in case but I think it should match all versions but mq-j

Yanis002: well, I'm confused, when I link for the other versions it acts the same as if it wasn't linked lol, but on objdiff it's matching, mq versions might require a dol-apply but that's it

cadmic: hm I built from clean and it worked

PR #203, opened 2025-04-12
Match _buildtev.c SetupStage

cadmic:
On mq-j, `lbl_801617B0` and friends (the frame temp buffers) are currently in _buildtev.c .bss but I think they probably make more sense in _frameGCNcc.c


Yanis002 (approved):
:tada:

PR #205, opened 2025-04-13
`simGCN.c` cleanup

Yanis002:
while working on MM I realised there was a lot of redundant code in this file so I made a macro for it, I don't think there's an "erased" function though, also this adds `simulatorDEMOSwapBuffers` and `simulatorDEMODoneRender` which are auto-inlines


cadmic (approved):
very nice!

PR #206, opened 2025-04-14
Match mcardGCN.c for mq-e

cadmic:
This wasn't that far off, but the diff looked bad because the functions were mixed around in symbols.txt


Yanis002 (approved):
nice!

PR #208, opened 2025-04-17
Start MM support

Yanis002:
This PR adds the config files and the basic changes to support MM, code changes will come in future PRs (there's a lot of things and I'd like to take the opportunity to do a self-review), I added the defines in `macros.h` (but I did not add things like `IS_MM` because I need to think more about this), also I updated the docs (note: I didn't add the badges in the readme because building doesn't work yet so it's not even supported in CI yet)

if you're wondering why I named it `mm-gc` (which gives the name to the final files) it's just that I find it weird to name that "oot-gc" when it's mm, but lmk if I should revert this change (including both bash script fixes), maybe we could give that a generic name?

if you want to see my draft branch (containing my current progress on the 3 regions): https://github.com/Yanis002/oot-gc/tree/mm-j


dol-apply
```diff
@@ -7,5 +7,11 @@
 set -euo pipefail
 
 VERSION=${1:-ce-j}
+if [ "${VERSION#mm-}" != "$VERSION" ]
```
cadmic, line 10: nit: `if [[ "$VERSION" = mm-* ]]` is probably clearer (also `[[` is generally better than `[`)

ditto in dol-diff

cadmic (approved):

PR #210, opened 2025-04-19
MM: match most of mcardGCN.c

Yanis002:
First big file, one notable change is `pMCard->saveToggle = false;` moved outside `mcardGameSetNoSave` as this function is not inlined on MM, otherwise it's line numbers changes and the new stuff, which wasn't cleaned up since I wasn't able to match this file unfortunately, there's only two non-matchings, it's `mcardWrite` and `mcardOpen`, I wasn't able to solve those

Another change is 4 new functions:
- `convert_i8i5`
- `convert_i5i8`
- `mcardSaveCamera`
- `mcardLoadZelda2Camera`


Yanis002:
also mm-u and mm-e matches less, I didn't spend a lot of time into it (tbh I forgot lol)

PR #211, opened 2025-04-19
MM: match most of simGCN.c

Yanis002:
This matches most of `simGCN.c`, the only functions left should be `simulatorDrawImage`, `simulatorDrawYesNoImage` and `simulatorDrawOKImage` (at least for mm-j and mm-u, I'm not gonna bother with mm-e on this file until it's matched for mq-e and ce-e), notable changes:
- TPL externs declarations moved to `THPRead.c` to workaround an issue on mm-e where it throws an error because the type became `u8*` (and non-simGCN files requires the array declaration to get a match unfortunately)
- align to 32 the TPL symbols
- moved `gContMap` and `gaszArgument` definitions to match .bss (because of static variables inside functions)
- changed the types of `gpFrame` and `gpSound` to their respective structs to match the debug data and also it was required for MM as it's accessing `gpFrame` (and I changed `gpSound` for consistency purposes)
- changed the `simulatorOpenMessage` macro to add another argument, on MM it does an `strcat` in there and it requires a new var called `buf` everywhere it's used
- removed `static` on `simulatorDEMODoneRender`, same thing it's required by MM as it's using that function in another file (and I thought it was pointless to ifdef that)
- fixed `gno` declared as `yno`

This is a lot of changes in a single file but I hope that's okay, after this file what's left to fix build issues are `system.c` and `_frameGCN.c`, then there's also `rom.c` and `library.c` to handle to get a matching build


config/mm-u/symbols.txt
```diff
@@ -1829,16 +1829,16 @@ gTypeFile = .data:0x800E0BC0; // type:object size:0x10 scope:global
...
+gcoverOpen = .data:0x800E0C60; // type:object size:0x2E61 scope:global align:32 noreloc
+gnoDisk = .data:0x800E3AE0; // type:object size:0x2501 scope:global align:32 noreloc
+gretryErr = .data:0x800E6000; // type:object size:0x3401 scope:global align:32 noreloc
+gfatalErr = .data:0x800E9420; // type:object size:0x4B41 scope:global align:32 noreloc
+gwrongDisk = .data:0x800EDF80; // type:object size:0x2501 scope:global align:32 noreloc
+greadingDisk = .data:0x800F04A0; // type:object size:0x1061 scope:global align:32 noreloc
+gbar = .data:0x800F1520; // type:object size:0x741 scope:global align:32 noreloc
+gyes = .data:0x800F1C80; // type:object size:0x141 scope:global align:32 noreloc
+gno = .data:0x800F1DE0; // type:object size:0x101 scope:global align:32 noreloc
+gmesgOK = .data:0x800F1F00; // type:object size:0x161 scope:global align:32 noreloc
```
cadmic, line 1841: huh, should we do this for all files for modding? like, do bad things happen if it's not aligned?
(hopefully it's just a regex find/replace)

src/emulator/simGCN.c
```diff
@@ -190,7 +217,11 @@ bool gPreviousIPLSetting;
 u32 gnTickReset;
 bool gbReset;
 
-#define simulatorOpenMessage(msg, fileName, size, fileInfo)                                     \
+#if IS_OOT
+#define PREFIX_TPL "TPL/"
```
cadmic, line 221: maybe TPL_DIR?
actually, do we need this at all? Maybe it can go into the `simulatorOpenMessage` directly. Not sure if that would make things confusing

src/emulator/simGCN.c
```diff
@@ -190,7 +217,11 @@ bool gPreviousIPLSetting;
...
+#define buf
```
cadmic, line 222: I'd find it surprising that something as innocent as `buf` is secretly a macro. Is it necessary? It seems like the `simulatorOpenMessage` macro ignores it anyway

src/emulator/simGCN.c
```diff
@@ -905,10 +1114,207 @@ bool simulatorDrawYesNoImage(TEXPalette* tplMessage, s32 nX0Message, s32 nY0Mess
...
+    GXSetTevColor(GX_TEVREG0, color);
+    GXSetTevColorOp(GX_TEVSTAGE0, GX_TEV_ADD, GX_TB_ZERO, GX_CS_SCALE_1, 0U, GX_TEVPREV);
+    GXSetTevAlphaOp(GX_TEVSTAGE0, GX_TEV_ADD, GX_TB_ZERO, GX_CS_SCALE_1, 0U, GX_TEVPREV);
+    GXSetTevColorIn(GX_TEVSTAGE0, GX_CC_ZERO, GX_CC_ZERO, GX_CC_ZERO, GX_CC_C0);
+    GXSetTevAlphaIn(GX_TEVSTAGE0, GX_CA_ZERO, GX_CA_ZERO, GX_CA_ZERO, GX_CA_KONST);
+    GXSetTevOrder(GX_TEVSTAGE0, GX_TEXCOORD_NULL, GX_TEXMAP_NULL, GX_COLOR_NULL);
+    GXSetBlendMode(GX_BM_NONE, GX_BL_SRCALPHA, GX_BL_INVSRCALPHA, GX_LO_NOOP);
+    GXClearVtxDesc();
+    GXSetVtxDesc(GX_VA_POS, GX_DIRECT);
+    GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_POS, GX_TEX_ST, GX_RGBA6, 0U);
```
cadmic, line 1153: GX_POS_XYZ, GX_F32

src/emulator/simGCN.c
```diff
@@ -1044,9 +1450,147 @@ bool simulatorDrawOKImage(TEXPalette* tplMessage, s32 nX0Message, s32 nY0Message
...
```
cadmic, line 1489: ditto

src/emulator/simGCN.c
```diff
@@ -190,7 +217,11 @@ bool gPreviousIPLSetting;
 u32 gnTickReset;
 bool gbReset;
 
-#define simulatorOpenMessage(msg, fileName, size, fileInfo)                                     \
+#if IS_OOT
+#define PREFIX_TPL "TPL/"
+#define buf
+
+#define simulatorOpenMessage(msg, fileName, size, fileInfo, ...)                                \
```
cadmic, line 224: nit: I think it's better to add an unused `buffer` argument instead of allowing this to take an arbitrary number of arguments

src/emulator/simGCN.c
```diff
@@ -1602,17 +2162,44 @@ bool simulatorReadController(s32 channel, u32* anData, u8* ptx) {
...
-        if (nCurrButton & PAD_BUTTON_LEFT) {
-            *anData |= gContMap[channel][GCN_BTN_DPAD_LEFT];
-        }
-        if (nCurrButton & PAD_BUTTON_RIGHT) {
-            *anData |= gContMap[channel][GCN_BTN_DPAD_RIGHT];
+
+#if IS_MM
+        if (gpSystem->eTypeROM == SRT_ZELDA2) {
+            if (nCurrButton & PAD_BUTTON_UP && !(nCurrButton & (PAD_TRIGGER_Z | PAD_BUTTON_A | PAD_BUTTON_X))) {
+                if (!(nDirButton & 0xF00)) {
```
cadmic, line 2169: 0xF00 -> `(GCN_BTN_DPAD_UP | GCN_BTN_DPAD_DOWN | GCN_BTN_DPAD_LEFT | GCN_BTN_DPAD_RIGHT)`?

src/emulator/simGCN.c
```diff
@@ -1938,6 +2528,249 @@ static bool simulatorDrawCursor(s32 nX, s32 nY) {
...
+    char path_yes[40];
+    char path_no[40];
+    char path_mesgOK[40];
+    struct DVDFileInfo fileInfo;
+    u32 size;
+
+    simulatorUnpackTexPalette((TEXPalette*)gbar);
+
+    switch (gLanguage) {
+        case 1:
```
cadmic, line 2548: hmm, eventually we should make an enum for this

cadmic (commented):
nice!

config/mm-u/symbols.txt
Yanis002: this was required to match .data sections, though it's odd because dtk used the alignment macro when defining the array in the .inc files

Yanis002: also for modding no idea, I think the TPLs are all aligned to 32

src/emulator/simGCN.c
Yanis002: I thought it was required to avoid errors but now that I think about it that's a macro so it should be fine

src/emulator/simGCN.c
Yanis002: yeah probably, I guess that will be part of simGCN cleanup once pal is matched

cadmic: from https://archive.org/stream/GCN_SDK_Documentation/Nintendo%20GameCube%20PAL%20Game%20Development%20Guidelines%20Version%200.2_djvu.txt apparently
```
#define OS_LANG_ENGLISH 0u // English
#define OS_LANG_GERMAN 1u // German
#define OS_LANG_FRENCH 2u // French
#define OS_LANG_SPANISH 3u // Spanish
#define OS_LANG_ITALIAN 4u // Italian
#define OS_LANG_DUTCH 5u // Dutch
```
(probably doesn't belong in this PR, I just happened to think about it)

cadmic (approved):

PR #213, opened 2025-04-19
MM: match system.c

Yanis002:
This is the last PR to fix the basic build issues (and also add the current progress on MM but there's few other files to complete that), it's matching `system.c` and `systemSetupGameALL` is annoying but that's it for the important changes

I mentioned this on discord few minutes ago but there's an interesting change in `contMap`, a 5th configuration was added, and as we now know on Wii VC there's also a 5th configuration (though they're not the same), I was already very suspicious of that but I think mm-j is the closest version we have to wii vc

Also last big pr for now, I wanted to address the build issues first, also feel free if you have suggestions to improve things in this file

### DIFF
--- a/src/emulator/system.c
+++ b/src/emulator/system.c
@@ -48,6 +48,7 @@
 #define MCARD_COMMENT "Zelda: Collector's Edition"
 #define MCARD_FILE_SIZE 0xC000
 #define MCARD_FILE_NAME_MM "ZELDA3"
+#define MCARD_COMMENT_MM "Legend of Zelda"
 #elif IS_MM_JP
 #define MCARD_FILE_NAME "ZELDA1"
 #define MCARD_FILE_NAME_MM "ZELDA2"
--- a/src/emulator/system.c
+++ b/src/emulator/system.c
@@ -29,17 +29,36 @@
 #define MCARD_FILE_NAME "ZELDA"
 #define MCARD_COMMENT "ゼルダの伝説　時のオカリナＧＣ" // "The Legend of Zelda: Ocarina of Time GC"
 #define MCARD_FILE_SIZE (0xC000 * 2)
+#define MCARD_FILE_NAME_MM "ZELDA3"
+#define MCARD_COMMENT_MM "Legend of Zelda"
 #elif VERSION == MQ_U || VERSION == MQ_E
 #define MCARD_FILE_NAME "ZELDA"
 #define MCARD_COMMENT "Zelda: Ocarina of Time"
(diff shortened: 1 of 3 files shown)


src/emulator/system.c
```diff
@@ -152,12 +197,48 @@ u32 nTickMultiplier = 2;
...
 #define Z_BNR_PATH "TPL/z_bnr.tpl"
+#define ROM_TEST_OOT_EU false
+#endif
+
+#if IS_MM_JP
+#define ROM_TEST_PW_JP romTestCode(pROM, "NPWJ")
+#define ROM_TEST_MM_US romTestCode(pROM, "NZSE")
+#define ROM_TEST_MM_EU romTestCode(pROM, "NZSP")
+#define ROM_TEST_MK64_JP romTestCode(pROM, "NKTJ")
+#define ROM_TEST_KIRBY_JP romTestCode(pROM, "NK4J")
```
cadmic, line 215: tbh I find these pretty artificial. I'd prefer either
```
if (
    romTestCode(pROM, "NZSJ") || romTestCode(pROM, "NZSE")
#if IS_MM_JP
    || romTestCode(pROM, "NZSP")
#endif
) {
    ...
}
```
or something like this
```
#if IS_MM_JP
#define ROM_IS_MM(pROM) (romTestCode(pROM, "NZSJ") || romTestCode(pROM, "NZSE") || romTestCode(pROM, "NZSP"))
#else 
#define ROM_IS_MM(pROM) (romTestCode(pROM, "NZSJ") || romTestCode(pROM, "NZSE"))
#endif
```

src/emulator/system.c
```diff
@@ -522,6 +646,15 @@ static bool systemSetupGameALL(System* pSystem) {
     } else if (gLanguage == 4) {
         strcpy(buf1, "TPL/ITALIAN/");
         strcpy(buf2, "TPL/ITALIAN/");
+#elif IS_MM_EU
+    } else if (gLanguage == 3) {
+        strcpy(buf1, "TPL/SPANISH/");
+        strcpy(buf2, "TPL/SPANISH/");
+    } else if (gLanguage == 3) {
+        //! @bug: This condition will always be false as `gLanguage` is 4 when using italian as the console's language.
+        //! This has no effect as this path is only used to get the memory card's icon and banner TPL files.
```
cadmic, line 655: Maybe something like "This code is never reached because the condition should be `gLanguage == 4` for Italian"?
Also why doesn't having a different memory card's icon and banner TPL count as an "effect"?

cadmic (approved):

PR #215, opened 2025-04-20
MM: match cpu.c

Yanis002:
Alright, this is the latest PR required to make MM build :ok: (after configure.py is updated of course), the diff is big because mm-j got a big and unused function (I have no idea why it's there), there's nothing that interesting to note, MM uses a register system (see `gRegCount` and `gRegList`) and there's `cpuPrintOpcode` but that's it

### DIFF
--- a/include/emulator/cpu.h
+++ b/include/emulator/cpu.h
@@ -19,6 +19,11 @@
 #define MIPS_IMM_U16(inst) ((u16)((inst) & 0xFFFF))
 #define MIPS_TARGET(inst) ((inst) & 0x3FFFFFF)
 
+//! TODO: fake match for `cpuPrintOpcode`
+#define MIPS_IMM_U16_2(inst) ((u16)((inst) & 0xFFFFF))
+
+#define MIPS_UNK(inst) ((inst) & ~0x3FF0000)
+
 #define MIPS_FMT(inst) (((inst) >> 21) & 0x1F)
 #define MIPS_FT(inst) (((inst) >> 16) & 0x1F)
 #define MIPS_FS(inst) (((inst) >> 11) & 0x1F)
--- a/include/emulator/cpu.h
+++ b/include/emulator/cpu.h
@@ -19,11 +19,6 @@
 #define MIPS_IMM_U16(inst) ((u16)((inst) & 0xFFFF))
 #define MIPS_TARGET(inst) ((inst) & 0x3FFFFFF)
 
-//! TODO: fake match for `cpuPrintOpcode`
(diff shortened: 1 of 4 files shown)


include/emulator/cpu.h
```diff
@@ -19,6 +19,11 @@
 #define MIPS_IMM_U16(inst) ((u16)((inst) & 0xFFFF))
 #define MIPS_TARGET(inst) ((inst) & 0x3FFFFFF)
 
+//! TODO: fake match for `cpuPrintOpcode`
+#define MIPS_IMM_U16_2(inst) ((u16)((inst) & 0xFFFFF))
+
+#define MIPS_UNK(inst) ((inst) & ~0x3FF0000)
```
cadmic, line 25: seems like this masks out `rs` and `rt`. Honestly I would just inline this instead of making it a macro, it seems like a hack

src/emulator/_cpuGCN.c
```diff
@@ -6479,1455 +6532,3303 @@ static inline bool cpuExecuteCacheInstruction(Cpu* pCPU) {
...
+                                strcpy(szText[0], "FLOOR.L");
+                                strcat(szText[0], ".S");
+                                strcpy(szText[1], gaszNameFPR[MIPS_SA(nOpcode)]);
+                                strcpy(szText[2], gaszNameFPR[MIPS_RD(nOpcode)]);
+                                break;
+                            case 0xC:
+                                strcpy(szText[0], "ROUND.W");
+                                strcat(szText[0], ".S");
+                                strcpy(szText[1], gaszNameFPR[MIPS_SA(nOpcode)]);
+                                strcpy(szText[2], gaszNameFPR[MIPS_RD(nOpcode)]);
```
cadmic, line 7269: sorry to make you do this but there's some confusion here about `MIPS_RS/MIPS_RT/MIPS_RD/MIPS_SA` and `MIPS_FMT/MIPS_FT/MIPS_FS/MIPS_FD` (the latter are for floating-point ops). Could you fix these? You can use cpuExecuteOpcode as a guide

include/emulator/cpu.h
Yanis002: I also removed `MIPS_IMM_U16_2` because I thought it was too weird since it's technically a fake match

src/emulator/_cpuGCN.c
Yanis002: I think I got them all but lmk if there's something wrong

cadmic (approved):

PR #216, opened 2025-04-25
Match `simGCN.c` for the remaining versions

Yanis002:
thanks to @cadmic's help, this file is matched on the europeans versions :tada:


src/emulator/simGCN.c
```diff
@@ -1591,70 +1599,158 @@ bool simulatorDrawOKImage(TEXPalette* tplMessage, s32 nX0Message, s32 nY0Message
 }
 #endif
 
-bool simulatorDrawErrorMessage(SimulatorMessage simulatorErrorMessage, bool drawBar, s32 percent) {
-    PADControlMotor(0, PAD_MOTOR_STOP);
-
+static inline void simulatorDrawErrorMessageImpl(SimulatorMessage simulatorErrorMessage, bool drawBar, s32 percent,
```
cadmic, line 1602: what about making this a macro too? This function doesn't seem to be real, and a macro would reduce some copy-paste and be more like the other functions

src/emulator/simGCN.c
```diff
@@ -1676,10 +1772,37 @@ bool simulatorPrepareMessage(SimulatorMessage simulatorErrorMessage) {
     return true;
 }
 
+#if IS_MM_EU
```
cadmic, line 1775: maybe it would be better to inline this version of the macro (like non-EU OOT)?

src/emulator/simGCN.c
```diff
@@ -1676,10 +1772,37 @@ bool simulatorPrepareMessage(SimulatorMessage simulatorErrorMessage) {
...
+                                180 - yes->descriptorArray->textureHeader->height / 2, no,                            \
+                                200 - no->descriptorArray->textureHeader->width / 2,                                  \
+                                180 - no->descriptorArray->textureHeader->height / 2);                                \
+    }
+#endif
+
 bool simulatorDrawYesNoMessageLoop(TEXPalette* simulatorQuestion, bool* yes) {
+#if IS_MM_EU
+    TEXDescriptor** descriptor;
+#elif VERSION != MQ_E && VERSION != CE_E
```
cadmic, line 1801: `!IS_OOT_EU`?

cadmic (approved):

PR #218, opened 2025-04-28
MM: build OK

Yanis002:
MM now builds :ok: and it was added to CI, there was tiny data issues I fixed (sbss order and a filename)


src/emulator/mcardGCN.c
```diff
@@ -2474,14 +2474,18 @@ bool mcardOpenDuringGameError(MemCard* pMCard, MemCardCommand* pCommand) {
 }
 
 #if IS_MM
-static bool mcardNewStart;
 
 #if IS_MM_JP
```
cadmic, line 2478: alternatively:
```
#if IS_MM_JP

#elif IS_MM // or, version is MM_US or MM_EU

#endif
```

cadmic (approved):

PR #219, opened 2025-04-29
MM: match `sndspecial.c`

Yanis002:
This file handles the credits screen that was (re)discovered 6 months ago. I tried a decent way to handle the strings from `ImportantData` but I couldn't find anything working unfortunately so instead I added comments

If you're wondering why that array looks like that, the reason is the format: the number is the length of the string to display, then it's the individual characters (I have no idea why they did it that way instead of using strings directly)


src/emulator/sndspecial.c
```diff
@@ -0,0 +1,308 @@
...
+    GXSetTevColorIn(GX_TEVSTAGE0, GX_CC_ZERO, GX_CC_TEXC, GX_CC_C0, GX_CC_ZERO);
+    GXSetTevAlphaIn(GX_TEVSTAGE0, GX_CA_ZERO, GX_CA_ZERO, GX_CA_ZERO, GX_CA_A0);
+    GXSetTevOrder(GX_TEVSTAGE0, GX_TEXCOORD0, GX_TEXMAP0, GX_COLOR_NULL);
+    GXSetBlendMode(GX_BM_BLEND, GX_BL_SRCALPHA, GX_BL_INVSRCALPHA, GX_LO_NOOP);
+    GXSetAlphaCompare(GX_ALWAYS, 0, GX_AOP_AND, GX_ALWAYS, 0);
+    GXSetZMode(GX_FALSE, GX_LEQUAL, GX_FALSE);
+    GXSetZCompLoc(GX_TRUE);
+    PSMTXIdentity(matrix);
+    GXLoadTexMtxImm(matrix, 0x1E, GX_MTX2x4);
+    GXInitTexObj(&texObj, special_data, 472, 8, GX_TF_I4, GX_CLAMP, GX_CLAMP, GX_FALSE);
```
cadmic, line 262: should we extract `special_data` as a texture like the others? (I think it's the font)

cadmic (approved):

cadmic:
also I'm pretty sure the weird data format and vague names is to hide this "feature" (from their bosses? lol)

src/emulator/sndspecial.c
Yanis002: yeah that makes sense

PR #220, opened 2025-05-01
Add progress categories

Yanis002:
also this updates the tools versions, now the output of `ninja` looks like this:
<details>

```
ce-j progress:
  All: 96.59% code, 97.91% data (159 / 181 files)
    Code: 822448 / 851456 bytes
    Data: 398068 / 406554 bytes
  Emulator: 100.00% code, 100.00% data (36 / 36 files)
    Code: 615708 / 615708 bytes
    Data: 372388 / 372388 bytes
```
</details>

I made it print only the "all" and "emulator" categories since that's what we care about (and having everything printed makes the output too big imo)


Yanis002:
the diff looks big because dtk wanted to update the config files

cadmic (approved):

PR #224, opened 2025-05-15
Match rspParseGBI_F3DEX{1,2}

cadmic:


src/emulator/rsp.c
```diff
@@ -5579,8 +5650,34 @@ bool rspBgRectCopy(Rsp* pRSP, Frame* pFrame, s32 nAddress) {
...
+    }
+
+    pnData16 = (u16*)pObjSubMtx;
+    nX = pnData16[0];
+    nY = pnData16[1];
+    nBaseScaleX = pnData16[2];
+    nBaseScaleY = pnData16[3];
+
+    pRSP->twoDValues.fX = (f32)nX / 4.0f;
+    pRSP->twoDValues.fY = -(f32)nY / 4.0f;
```
Yanis002, line 5672: it seems the `nY` cast is the only one necessary to match the function, but not sure how it affects the data

Yanis002 (approved):

src/emulator/rsp.c
cadmic: Without an explicit `(f32)`, the other ones would get casted to a float anyway since it's getting divided by a float, so it won't affect the data at all. The `(f32)` on nY is necessary though, because without it `-nY / 4.0f` will negate first as an integer and then cast to a float. So I put in the others for consistency

PR #225, opened 2025-05-15
Match S2DEX helper functions

cadmic:


src/emulator/rsp.c
```diff
@@ -5526,9 +5633,398 @@ static bool rspObjRectangle(Rsp* pRSP, Frame* pFrame, s32 nAddress) {
...
+    s32 nClampSetting;
+    uObjSprite objSprite;
+    Tile* pTile;
+    s32 pad[2];
+    Primitive primitive;
+    f32 mtxTransL[3][4];
+    f32 mtxTransW[3][4];
+    f32 mtxScale[3][4];
+    f32 mtxTemp[3][4];
+    f32 mtxOut[3][4];
```
Yanis002, line 5666: these should be `Mtx` instead of `f32[3][4]`

src/emulator/rsp.c
```diff
@@ -5526,9 +5633,398 @@ static bool rspObjRectangle(Rsp* pRSP, Frame* pFrame, s32 nAddress) {
...
+    s32 nClampSetting;
+    uObjSprite objSprite;
+    s32 pad2[2];
+    Tile* pTile;
+    Primitive primitive;
+    f32 mtxTransL[3][4];
+    f32 mtxTransW[3][4];
+    f32 mtxScale[3][4];
+    f32 mtxTemp[3][4];
+    f32 mtxOut[3][4];
```
Yanis002, line 5865: `Mtx`

src/emulator/rsp.c
cadmic: good catch thanks

PR #227, opened 2025-05-16
Dolphin SDK headers cleanup

Yanis002:
This PR syncs the dolphin SDK headers with [dolsdk2004](https://github.com/doldecomp/dolsdk2004), also this matches `GXLight.c` for MQ (side effect of cleaning up GX headers), changes in source files are necessary changes to fix build and non-matching issues


include/cmath.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _DOLPHIN_CMATH_H_
+#define _DOLPHIN_CMATH_H_
+
+float powf(float x, float y);
+float tanf(float);
+
+extern float sinf(float);
+extern float cosf(float);
+extern float acosf(float);
+extern float atan2f(float, float);
```
cadmic, line 10: how does this relate to libc/math.h? I notice that also declares these functions

cadmic (approved):

include/cmath.h
Yanis002: I moved the missing declarations to math.h and deleted this file

PR #229, opened 2025-05-18
Cleanup 1

Yanis002:
my gcc experiments revealed changes required to build properly with it, this is what this PR brings (basically this is all missing casts, mostly `(void**)`), also I added a pragma around `cpuGetPPC` explaining why this is useful (tldr good for modding but this compiler version doesn't understand it so it has no effect on matching builds)

### DIFF
--- a/src/emulator/_cpuGCN.c
+++ b/src/emulator/_cpuGCN.c
@@ -227,6 +227,10 @@ static inline void cpuCompileNOP(s32* anCode, s32* iCode, s32 number) {
         }                              \
     } while (0)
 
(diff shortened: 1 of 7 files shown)


src/emulator/_cpuGCN.c
```diff
@@ -227,6 +227,10 @@ static inline void cpuCompileNOP(s32* anCode, s32* iCode, s32 number) {
         }                              \
     } while (0)
 
+// Note: the original compiler (GC/1.1) doesn't understand this but it's useful for modding as it helps reducing the
```
cadmic, line 230: reducing -> reduce

cadmic (approved):
looks good after format

PR #240, opened 2025-05-22
Delete asm_processor

cadmic:
It's no longer used and will probably never be used again, and deleting it will make merging project.py (or upstreaming the multi-version support) easier


cadmic:
(depends on #239)

Yanis002 (approved):
post hundo cleanup feels nice :broom:

PR #241, opened 2025-05-22
Delete checked-in disassembly

cadmic:
Really this was only useful for asm_processor since dtk's own disassembly doesn't play nice with it


Yanis002 (approved):
btw I think you can delete the checksum file at the root, I noticed this recently it's probably a left-over from the older build system

PR #242, opened 2025-05-22
Clean up instances of if (cond && !func())

cadmic:
for instances of 
```
if (cond && !func(...)) {
    return false;
}
```
where `cond` is a "check something" thing and `func` is a "do something" thing, rewrite this as
```
if (cond) {
    if (!func(...)) {
        return false;
    }
}
```
I find this easier to read since the `if (!func(...)) return false;` is standard boilerplate and isn't really part of the condition.

### DIFF
--- a/src/emulator/simGCN.c
+++ b/src/emulator/simGCN.c
@@ -2781,165 +2781,164 @@ bool simulatorPreloadDiskMessages(void) {
     if (DVDOpen(path_readingDisk, &fileInfo)) {
         size = (fileInfo.length + 0x1F) & ~0x1F;
 
-        if (greadingDisk != NULL)
+        if (greadingDisk != NULL) {
             if (!xlHeapFree((void**)&greadingDisk)) {
                 return false;
             }
-    }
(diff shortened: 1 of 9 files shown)


src/emulator/simGCN.c
```diff
@@ -2777,148 +2781,165 @@ bool simulatorPreloadDiskMessages(void) {
...
 
-    return true;
+    simulatorDVDRead(&fileInfo, gmesgOK, size, 0, NULL);
+    DVDClose(&fileInfo);
+    simulatorUnpackTexPalette((TEXPalette*)gmesgOK);
+} else {
+    return false;
+}
+
+return true;
```
Yanis002, line 2942: what's up with the formatting here?

cadmic: uhh I probably broke something. Will fix soon

cadmic: should be fixed now

Yanis002 (approved):

PR #243, opened 2025-05-22
Mark pFrame->nMode and pSound->eMode as volatile

cadmic:
These are probably not fake


Yanis002 (approved):
wii vc might help on solving things like this

